PR #64, opened 2017-03-28
Allow to create requested secret from mode and path

pvardanega:
This change gives the user the opportunity to create a requested secret from a mode that is injected from somewhere, like configuration file for instance (see [example](https://github.com/spring-cloud/spring-cloud-vault-config/pull/90)). It also avoids the user to check the mode before creating it each time.

Regarding OkHttp support removal. It seems that Spring 5 has removed its support. The easiest solution for me was to remove it also but I don't know the policy. If you want to keep it, we'll need to either rollback this [commit](https://github.com/spring-projects/spring-framework/commit/47c4cf7abf231018e102bf51d83dd92c656f8996) or move the removed classes somewhere else.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/RequestedSecret.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/RequestedSecret.java
@@ -17,6 +17,9 @@ package org.springframework.vault.core.lease.domain;
 
 import org.springframework.util.Assert;
 
+import static org.springframework.vault.core.lease.domain.RequestedSecret.Mode.RENEW;
+import static org.springframework.vault.core.lease.domain.RequestedSecret.Mode.ROTATE;
+
 /**
  * Represents a requested secret from a specific Vault path associated with a lease
  * {@link Mode}.
@@ -50,7 +53,7 @@ public class RequestedSecret {
 	 * @return the renewable {@link RequestedSecret}.
 	 */
 	public static RequestedSecret renewable(String path) {
-		return new RequestedSecret(path, Mode.RENEW);
+		return new RequestedSecret(path, RENEW);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/RequestedSecret.java
```diff
@@ -95,12 +109,24 @@ public String toString() {
...
 		 */
-		ROTATE;
+		ROTATE {
+			@Override
+			public RequestedSecret buildRequestedSecret(String path) {
+				return rotating(path);
+			}
+		};
+
+		abstract RequestedSecret buildRequestedSecret(String path);
```
mp911de: Introduction of the new API methods creates a cycle to `RequestedSecret`. How about keeping the whole construction inside `buildFromMode(…)`?

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/RequestedSecret.java
```diff
@@ -66,6 +66,20 @@ public static RequestedSecret rotating(String path) {
...
+	 * Create a rotating or renewable {@link RequestedSecret} at {@code path}. A lease associated with
+	 * this secret will be renewed if the lease is qualified for renewal. Once the lease
+	 * expires, a new secret with a new lease is obtained if mode is ROTATE, otherwize the lease is no
+	 * longer valid after expiry.
+	 *
+	 * @param mode must not be {@literal null}
+	 * @param path must not be {@literal null} or empty, must not start with a slash.
+	 * @return the rotating {@link RequestedSecret}.
+	 */
+	public static RequestedSecret buildFromMode(Mode mode, String path) {
```
mp911de: How about just `from(Mode, Path)`?

I'd propose to use a simple `if` construction instead of introducing API methods on `Mode`. You might want to `Assert.notNull(mode, …)` here as well.

pvardanega: Done

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/RequestedSecret.java
pvardanega: Yep, done

pvardanega:
Thank you for your review. I've applied your comments.

I did not know how to deal with spring5 compatibility so I'll revert the commit. Is it ok if I squash the 2 commits in the previous and push force on the branch? Or do you prefer to keep the commit + revert on the commit tree?

mp911de:
Thanks a lot. I'll take the PR from here and take care of squashing

mp911de:
That's merged and polished now.

Issue #68, opened 2017-04-04
Add automatic rotation of secrets stored in generic backend


deichten:
At the moment the renewal process is dependent on a lease_id and evaluating if a lease is renewable which is not the case for secrets stored in the generic backend which is by design not returning any leases. As indicated in the [documentation of Vault](https://www.vaultproject.io/docs/secrets/generic/index.html#quick-start) the client is supposed to check back the value after the `refresh_interval` expired. I was stumbling over this implementing some test code. I suggest to do a quick fix in the documentation as well as potentially add a new Renewal strategy, e.g. `REFRESH`, to have the `SecretLeaseContainer` schedule the refresh of the secrets retrieved.

mp911de:
Added a note about the current behavior in the reference documentation for now. 

It makes sense to refresh generic secrets (if requested with `Renewal.ROTATE`) and update the `PropertySource` content. I'm not quite sure whether it makes sense to remove these secrets in an error case (token is no longer authorized to obtain the secrets (HTTP 403), secrets removed by an operator (HTTP 404)). Happy to receive feedback.

crinacimpian:
When will there be a fix for "refresh generic secrets (if requested with Renewal.ROTATE) and update the PropertySource content"? And my feedback: remove secrets from PropertySource content on 403/ 404 error cases.

mp911de:
This ticket is likely to be implemented for Spring Vault 1.1. Feel free to submit a PR.

sworisbreathing:
Per https://www.vaultproject.io/api/secret/generic/index.html:

> it is possible to mount secret backends at any location

So to implement this we'll also need to expose some configuration properties to tell Spring Vault which paths map to generic secret backends. [Spring Cloud Vault supports this](https://github.com/spring-cloud/spring-cloud-vault/blob/v1.0.0.RELEASE/spring-cloud-vault-config/src/main/java/org/springframework/cloud/vault/config/VaultGenericBackendProperties.java#L47) but it appears to only support a single generic backend.

sworisbreathing:
Also, we can't use the lease renewal api in vault for this, because it's not a renewable lease. So we'd need a separate mechanism to schedule reloading of generic secrets based on `lease_duration` (assuming it's > 0).

mp911de:
I adjusted the title to secret rotation for the generic secret backend. The term `renewal` has in the Vault space a different meaning and since there's no lease associated with generic secrets lease renewal cannot be applied.

[...]

PR #95, opened 2017-05-15
Rotate generic secrets based on lease_duration/ttl. Fixes gh-68

sworisbreathing:


pivotal-issuemaster:
@sworisbreathing Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=95)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-vault&pullRequestId=95) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sworisbreathing Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=95)!

sworisbreathing:
Grrr... two of the travis build jobs failed, but I don't understand why, and the logs seem to have disappeared. @mp911de - can you maybe point me in the right direction here?

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/Lease.java
```diff
@@ -42,14 +44,14 @@ private Lease(String leaseId, long leaseDuration, boolean renewable) {
 	/**
 	 * Create a new {@link Lease}.
 	 *
-	 * @param leaseId must not be empty or {@literal null}.
+	 * @param leaseId must not {@literal null}. Empty string implies a generic secret.
```
mp911de: I would like to keep this method as-is and rather introduce a new factory method just accepting `leaseDuration`. I'm not decided on a name yet, something expressing the source as cache duration hint (`fromTtl`, `fromCacheHint`, …) might be a good fit. WDYT?

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/Lease.java
```diff
@@ -84,6 +86,14 @@ public boolean isRenewable() {
 		return renewable;
 	}
 
+	/**
+	 *
+	 * @return {@literal true} if the lease represents a rotating generic secret.
+	 */
+	public boolean isRotatingGenericLease() {
```
mp911de: Rotation is a Spring Vault concept, not necessarily bound to a `Lease`. I'd propose to rather introduce a method `boolean hasLeaseId()` instead. Rotation and generic secret backend aren't really tied to a lease.

spring-vault-core/src/test/java/org/springframework/vault/core/lease/SecretLeaseContainerIntegrationTests.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Integration tests for {@link SecretLeaseContainer} using the {@code generic} backend.
+ *
+ * @author Steven Swor
+ */
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = { VaultIntegrationTestConfiguration.class,
+		RotatingGenericSecretsIntegrationTestConfiguration.class })
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+public class SecretLeaseContainerIntegrationTests {
```
mp911de: Let's remove this integration test to keep the number of integration tests low. We have already unit tests for `SecretLeaseContainer` with renewal in `SecretLeaseContainerUnitTests` so the new test case could fit in there quite well.

Especially the timing aspect makes testing inconvenient because it adds a dependency on something we don't have under control.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
```diff
@@ -375,7 +385,7 @@ public void destroy() throws Exception {
 					Lease lease = entry.getValue().getLease();
 					entry.getValue().disableScheduleRenewal();
 
-					if (lease != null) {
+					if (lease != null && !lease.isRotatingGenericLease()) {
```
mp911de: `Lease.hasLeaseId()` would be a good candidate here

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
```diff
@@ -290,16 +291,25 @@ public void start() {
 		}
 	}
 
+	private boolean isRotatingGenericSecret(RequestedSecret requestedSecret,
```
mp911de: method can be static.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
```diff
@@ -463,10 +474,15 @@ protected Lease doRenewLease(RequestedSecret requestedSecret, final Lease lease)
 						@Override
 						@SuppressWarnings("unchecked")
 						public ResponseEntity<Map<String, Object>> doWithRestOperations(
-								RestOperations restOperations) {
+										RestOperations restOperations) {
```
mp911de: It could make rather sense to use `onLeaseExpired` instead of adding conditions to handle generic secrets. `onLeaseExpired` will handle rotation (request new secrets) already. 

`LeaseAwareVaultPropertySource` clears right now the contents of the property source on expiry to prevent stale data.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
sworisbreathing: There's a bit more logic to `isRotatingGenericLease`:

1. empty/null lease ID
1. non-renewable
1. positive lease duration

The first two give you some pretty good confidence you're looking at a generic secret. The last one tells you whether or not you should expect it to rotate.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
sworisbreathing: That makes sense I suppose, but I'm not sure how to implement it.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/Lease.java
sworisbreathing: the parameters are all of different types, so I don't see the harm in reusing `of`. will push this change up shortly

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/Lease.java
sworisbreathing: Perhaps, but it was a convenient place to stick the logic.

spring-vault-core/src/test/java/org/springframework/vault/core/lease/SecretLeaseContainerIntegrationTests.java
sworisbreathing: I'll `@Ignore` it for now. I'm hesitant to remove it outright

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
mp911de: Leases without a lease Id aren't required to be revoked because there's no Lease Id. `Lease.none` is as well a candidate that does not require a revocation. Call it a bug fix to be truly honest.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
mp911de: It should be sufficient to return just `null` for a `Lease` without `leaseId` and remove your additions. Everything else should happen itself.

spring-vault-core/src/main/java/org/springframework/vault/core/lease/domain/Lease.java
mp911de: So let's go ahead with the `hasLeaseId` method.

mp911de:
I'll take the pull request from here.

mp911de:
That's squashed, merged and polished. Thanks a lot for your great contribution.

Issue #96, opened 2017-05-15
Login Token Duration is assumed milliseconds instead of seconds


sworisbreathing:
Per https://www.vaultproject.io/docs/concepts/lease.html#lease-durations-and-renewal:

>The lease duration is a Time To Live value: the time in seconds for which the lease is valid. A consumer of this secret must renew the lease within that time.

However it looks like LifecycleAwareSessionManager [assumes lease durations are in milliseconds](https://github.com/spring-projects/spring-vault/blob/e428cbf846e7e91b8ae30fc7575c306ae2e1848d/spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java#L324)

To verify, run a simple client app with `-Dorg.apache.http.wire=debug` and watch your login token get renewed every second.

zonybob:
Any ETA for 1.0.2 release? Thanks!

mp911de:
I updated the schedule for [1.0.2](https://github.com/spring-projects/spring-vault/milestone/11) to June 16th.

PR #97, opened 2017-05-15
Convert login token duration to milliseconds when calculating renewal…

sworisbreathing:
… time. Fixes gh-96 (#96)


mp911de:
Thanks for your PR. I will merge it once I'm back from vacation (next week).

mp911de:
That's merged, polished and ported to 1.0.x.

PR #100, opened 2017-05-19
Allow health check to pass `standbyok` parameter.

sworisbreathing:
Beginning with Vault 0.2.0, the `/sys/health` endpoint for Vault allows an optional request parameter, `standbyok`, which, when set to `true` (default is `false`) will cause the health check to return a 200 status instead of a 429 status for a standby Vault node. It would be nice if the health check in Spring Vault supported this feature.

See https://www.vaultproject.io/api/system/health.html

Refs spring-cloud/spring-cloud-vault#112

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultSysOperations.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultSysOperations.java
@@ -127,11 +127,25 @@ public interface VaultSysOperations {
 	void authUnmount(String path) throws VaultException;
 
 	/**
-	 * Return the health status of Vault.
+	 * Return the health status of Vault. Implementations should ensure this is
+	 * functionally equivalent to {@link #health(boolean) health(false)}.
 	 *
 	 * @return the {@link VaultHealth}.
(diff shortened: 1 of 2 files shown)


mp911de:
I'm not sure what you're trying to achieve with the change. `VaultSysOperations.health()` returns in every case a `VaultHealth` object with the fields in the response set (see [`VaultSysTemplate.java`](https://github.com/spring-projects/spring-vault/blob/a9c2cf72f824857ff3c1654fdc0cf5987889ae16/spring-vault-core/src/main/java/org/springframework/vault/core/VaultSysTemplate.java#L298-L310)).

I'll add a comment to spring-cloud/spring-cloud-vault#112 which does not depend on changes in Spring Vault.

mp911de:
Btw. congrats, you hit issue number 💯

sworisbreathing:
Hrm... good point. Spring Cloud Vault doesn't need to rely on the parameter passed in to the API request, to alter its behavior. There might be an edge case somewhere, where the status code would matter, but I can't think of a good example.

mp911de:
Thanks for your feedback. I'm closing this PR without a merge.

Issue #132, opened 2017-08-31
Support AppRole pull mode


mp911de:
Add support for pull mode by fetching a wrapped `secret_id` with an initial token.

See also: [“Secret id missing” error while connecting to Vault using Spring cloud vault]( https://stackoverflow.com/questions/45978288/secret-id-missing-error-while-connecting-to-vault-using-spring-cloud-vault/45983670#45983670).

vincent-lenair:
Hi,

I am facing the same issue, I need to use pull mode to get RoleID and SecretID.
I am planning to make changes in AppRoleAuthentication to handle Push/Pull mode, introducing 2 new parameters: AppRole and RoleToken that would allow the application to use pull mode in order to get RoleID and/or SecretID

The logic would be the following:

```
IF RoleID AND SecretID, use push mode (current behavior)
ELSE IF AppRole AND RoleToken
        IF NOT RoleID, pull RoleID from Vault API using AppRole and RoleToken
        IF NOT SecretID, pull SecretID from Vault API using AppRole and RoleToken
```

then authenticate using RoleID and SecretID to get secrets as usual

mp911de:
Thanks for your comment. I'm a bit torn on the implementation. In general, your approach would work but the thing is, that pull mode requires additional context and would incorporate I/O. 

I'm thinking about another AppRole authentication implementation class along with separate options that allow pulling a roleId and secretId from either wrapped responses or the authentication backend itself.

I would describe the options with `VaultToken` and a `Source` object for RoleId and SecretId so the actual `ClientAuthentication` can then decide where to fetch the data from, like you described it above.

vincent-lenair:
Hi,

Thanks for your reply.

Can you explain what you mean by "pull mode requires additional context and would incorporate I/O"
I haven't started working on it, but I by looking at it I though I would be able to get the role ID and secret ID using RestOperations, and then get the VaultToken in the same way as we do in AppRoleAuthentication.createTokenUsingAppRole

I can implement a new type of authentication, something like a RoleNameAuthentication, using the AppRole and RoleToken in order to get the RoleID and SecretID using the pull mode but I think this solution is less generic.

Not sure this is what you had in mind

mp911de:
Pull mode fetches at least SecretId from Vault – either from a wrapped response or from `approle/role/(role)/secret-id`. Compared to `AppRoleAuthentication`, that's an additional HTTP request to be made. I guess it makes sense to provide something similar for RoleId.

[...]

PR #133, opened 2017-09-06
Allow Pull Mode when using the AppRoleAuthentication by sending on Ap…

vincent-lenair:
…pRole and RoleToken.

Fixes gh-132

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
@@ -17,11 +17,14 @@ package org.springframework.vault.authentication;
 
 import java.util.HashMap;
 import java.util.Map;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.vault.VaultException;
 import org.springframework.vault.client.VaultResponses;
 import org.springframework.vault.support.VaultResponse;
@@ -74,8 +77,10 @@ public class AppRoleAuthentication implements ClientAuthentication {
 
 	private VaultToken createTokenUsingAppRole() {
 
-		Map<String, String> login = getAppRoleLogin(options.getRoleId(),
-				options.getSecretId());
+		String roleId = getRoleId();
+		String secretId = getSecretId();
+
+		Map<String, String> login = getAppRoleLogin(roleId, secretId);
 
 		try {
 			VaultResponse response = restOperations.postForObject("auth/{mount}/login",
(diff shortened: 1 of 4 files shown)


spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthenticationOptions.java
```diff
@@ -148,9 +207,16 @@ public AppRoleAuthenticationOptionsBuilder secretId(String secretId) {
 		public AppRoleAuthenticationOptions build() {
 
 			Assert.hasText(path, "Path must not be empty");
-			Assert.notNull(roleId, "RoleId must not be null");
 
-			return new AppRoleAuthenticationOptions(path, roleId, secretId);
+			//Need to have either RoleID or (AppRole and RoleToken)
+			if(StringUtils.isEmpty(appRole) && StringUtils.isEmpty(roleToken)){
```
mp911de: How about splitting these `if`s and decouple the cases? `if`...`else` is quite hard to read. We should also verify the various combinations (roleId/secretId are both present -> token/role name not required, roleId or secretId missing -> token/role must be present).

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthenticationOptions.java
```diff
@@ -112,6 +143,34 @@ public AppRoleAuthenticationOptionsBuilder path(String path) {
...
+			return this;
+		}
+
+		/**
+		 * Configure a {@code roleToken}.
+		 *
+		 * @param roleToken must not be empty or {@literal null}.
+		 * @return {@code this} {@link AppRoleAuthenticationOptionsBuilder}.
+		 */
+		public AppRoleAuthenticationOptionsBuilder roleToken(String roleToken) {
```
mp911de: Let's rename the token to `initialToken`, like it's named in `CubbyholeAuthentication`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -91,6 +96,48 @@ private VaultToken createTokenUsingAppRole() {
 		}
 	}
 
+	private String getRoleId() {
+		String roleId = options.getRoleId();
+		if (StringUtils.isEmpty(roleId) && !StringUtils.isEmpty(options.getAppRole())) {
```
mp911de: We should assume that `appRole` is set otherwise this case could return an empty RoleId.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -91,6 +96,48 @@ private VaultToken createTokenUsingAppRole() {
...
+					VaultResponses.getError(e.getResponseBodyAsString())));
+			}
+		}
+		return roleId;
+
+	}
+
+	private String getSecretId() {
+		String secretId = options.getSecretId();
+		if(StringUtils.isEmpty(secretId) && !StringUtils.isEmpty(options.getAppRole())) {
```
mp911de: We should assume that `appRole` is set otherwise this case could return an empty SecretId.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthenticationOptions.java
vincent-lenair: I'll try to make it easier to read.
Checking the actual implementation, secretId can be ignored if roleId is set.
The combinations I am thinking about are:
- appRole and initialToken are empty -> we need a roleId to authenticate
- roleId is empty -> I need a appRole and a token to get at least the roleId

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
vincent-lenair: Does that mean I should remove the appRole verification there and rely on the Assertions from AppRoleAuthenticationOptions ?

mp911de: Yes. We expect a valid options object when constructing an `AppRoleAuthentication` instance.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthenticationOptions.java
mp911de: You're right, secretId is optional – When it's empty and a token is present then it's the pull case to fetch the secretId before authentication.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthenticationOptions.java
vincent-lenair: Done

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthenticationOptions.java
vincent-lenair: I have changed the initial check to reflect this logic

mp911de:
Thanks a lot for the updates. I'll take the PR from here.

mp911de:
That's merged, polished and ported to 2.0.x now. Thanks a lot.

Issue #137, opened 2017-09-12
Add support for Batch Encryption/Decryption APIs


pravsingh:
Vault supports the batch encryption & decryption services. Which is mostly used to achieve performance.

ref: [Batch Transit Operations](https://www.vaultproject.io/api/secret/transit/index.html#batch_input)

I'm working on these APIs for my use case and when testing is complete will create PR from my fork https://github.com/pravsingh/spring-vault.

Please plan this ticket based on the release schedule.

Thanks.

mp911de:
Spring Vault is currently in the release candidate phase and we shouldn't add any changes that could impact core functionality. Please see our [milestones](https://github.com/spring-projects/spring-vault/milestone/9) for release dates. 

Batch encryption seems not to entirely impact Spring Vault, only the transit API. I'd like to include the change if you want to submit a pull request within the next days.

pravsingh:
thanks for the guidance @mp911de. I'll try to send the PR today.

pravsingh:
created https://github.com/spring-projects/spring-vault/pull/138

pravsingh:
could you please add this ticket to 1.1.0 GA milestone?

pravsingh:
thank you.

pravsingh:
@mp911de,
any proposed date for the next release 1.1.0?
I'm eagerly waiting for the release to utilize the batch APIs in my product.
thanks

mp911de:
Release dates are documented in the milestones: https://github.com/spring-projects/spring-vault/milestones. 1.1.0 GA will ship after the Spring Framework 4.3.12 release.

pravsingh:
awesome. thanks.

PR #138, opened 2017-09-13
Added transit batch encrypt and decrypt support

pravsingh:
Per [https://github.com/spring-projects/spring-vault/issues/137](issue#137), added encrypt & decrypt APIs which does utilizes the batch operation of vault transit backend.

 please review and let me know if there is any room for improvement.

+ @mp911de


README.adoc
```diff
@@ -246,6 +246,7 @@ The following scripts need to be run prior to building the project for the tests
 
     $ ./src/test/bash/install_vault.sh
     $ ./src/test/bash/create_certificates.sh
+    $ ./src/test/bash/env.sh
```
pravsingh: I had to run this script to make local build work with vault.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -123,6 +124,16 @@
 	String encrypt(String keyName, byte[] plaintext, VaultTransitContext transitRequest);
 
 	/**
```
pravsingh: these are the first version of encrypt/decrypt APIs which accept the List of String. However we can add List/Array of byte[] version of the APIs and also the one which takes VaultTransitContext.
Once this PR is accepted, I can create ticket for these flavors of the APIs and work on for the next snapshot version.

let me know your thought.
thanks.

src/main/asciidoc/new-features.adoc
```diff
@@ -5,3 +5,8 @@
 === What's new in Spring Vault 1.0
 
 * Initial Vault support.
+
+[[new-features.1-1-0]]
+=== What's new in Spring Vault 1.1.0
+
+* Batch encryption & decryption support.
```
pravsingh: let me know if this is proper location or if you are managing the feature list somewhere else.

pravsingh:
need quick help on understanding why build failed on Travis CI however passes on local.

error:
```text
org.springframework.vault.VaultException: Status 400 transit/encrypt/mykey: missing plaintext to encrypt

	at org.springframework.vault.core.VaultTransitTemplateIntegrationTests.batchEncryptionAndDecryption(VaultTransitTemplateIntegrationTests.java:294)
```

pravsingh:
surprisingly travis is not using 0.8.2 version of vault which is defined in .travis.yml:
```bash
[33;1mSetting environment variables from .travis.yml[0m
$ export VAULT_VER=0.5.2
```

pravsingh:
I was able to reproduce the issue on my local
by setting 
```bash VAULT_VER="${VAULT_VER:-0.5.2}" 
```
in src/test/bash/install_vault.sh and installing the vault again and running the test cases.

Looks like the batch transit operations were not available in 0.5.2.
Tested and found out that the earliest version when batch operation worked was in 0.6.5.
Hence we need at least 0.6.5 to pass this PR.

pravsingh:
6 variations of the build failed due to following iterations:
```bash
    - VAULT_VER=0.5.2
    - VAULT_VER=0.6.0
    - VAULT_VER=0.6.1
    - VAULT_VER=0.6.2
    - VAULT_VER=0.6.3
    - VAULT_VER=0.6.4
```

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -123,6 +124,32 @@
...
+	 * Encrypts the provided list of plaintext using the named key.
+	 * The encryption is done using transit backend's batch operation.
+	 *
+	 * works with Vault 0.6.5 and later.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plaintexts a list of non-empty (or {@literal null}) plaintext.
+	 * @return list of cipher text in the same order as in plaintexts.
+	 */
+	List<Map<String, String>> encrypt(String keyName, List<String> plaintexts);
```
mp911de: Returning a `Map` of `String`s isn't quite expressive. It's not obvious what the key and what the values are. I'd rather return a `List<String>`. You give the method an ordered list of plaintexts and receive an ordered list of ciphertexts.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -123,6 +124,32 @@
...
+	 *
+	 * works with Vault 0.6.5 and later.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plaintexts a list of non-empty (or {@literal null}) plaintext.
+	 * @param transitRequests a list of  VaultTransitContext. May be {@literal null} if no request options provided.
+	 * @return list of cipher text in the same order as in plaintexts.
+	 * @throws IllegalArgumentException in case of not matching context found.
+	 */
+	List<Map<String, String>> encrypt(String keyName, List<String> plaintexts, List<VaultTransitContext> transitRequests);
```
mp911de: Same here as above. Additionally, the two lists are not connected – these are just two lists that have no relation to each other. With the single-operation API it wasn't quite obvious but plaintext + context/ciphertext + context have a relation which shouldn't be split up. Maybe we could introduce value objects (`Plaintext`, `Ciphertext`) which encapsulate the actual content along the optional context.

Returning a `List<Ciphertext>` or `List<Encrypted>` (the method before that one was `List<String>`) would express a slightly different context as the caller passes a `List<Plaintext>` to this method. The result could also encapsulate the success/error state of each operation.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -142,6 +169,32 @@
...
+	 * Decrypts the provided list of ciphertext using the named key.
+	 * The decryption is done using transit backend's batch operation.
+	 * 
+	 * works with Vault 0.6.5 and later.
+	 * 
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param ciphertexts a list of non-empty (or {@literal null}) ciphertext.
+	 * @return list of plain text in the same order as in ciphertexts.
+	 */
+	List<Map<String, String>> decrypt(String keyName, List<String> ciphertexts);
```
mp911de: See above.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -142,6 +169,32 @@
...
+	 * 
+	 * works with Vault 0.6.5 and later.
+	 * 
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param ciphertexts a list of non-empty (or {@literal null}) ciphertext.
+	 * @param transitRequests a list of  VaultTransitContext. May be {@literal null} if no request options provided.
+	 * @return list of plain text in the same order as in ciphertexts.
+	 * @throws IllegalArgumentException in case of not matching context found.
+	 */
+	List<Map<String, String>> decrypt(String keyName, List<String> ciphertexts, List<VaultTransitContext> transitRequests);
```
mp911de: See above.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *	  http://www.apache.org/licenses/LICENSE-2.0
```
mp911de: Please do not alter the license headers.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -174,6 +178,49 @@ public String encrypt(String keyName, byte[] plaintext,
...
+		return encrypt(keyName, plaintexts, null);
+	}
+	
+	@Override
+	public List<Map<String, String>> encrypt(String keyName, List<String> plaintexts,
+			List<VaultTransitContext> transitRequests) {
+		
+		Boolean hasContext = false;
+		
+		if(transitRequests!= null) {
```
mp911de: Coupling the actual plaintext to the context via a `Plaintext` object simplifies the implementation and is actually easier to comprehend from the usage side.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -174,6 +178,49 @@ public String encrypt(String keyName, byte[] plaintext,
...
+	public List<Map<String, String>> encrypt(String keyName, List<String> plaintexts) {
+
+		return encrypt(keyName, plaintexts, null);
+	}
+	
+	@Override
+	public List<Map<String, String>> encrypt(String keyName, List<String> plaintexts,
+			List<VaultTransitContext> transitRequests) {
+		
+		Boolean hasContext = false;
```
mp911de: How about using the primitive `boolean`? There's not need for three states here.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -211,6 +258,50 @@ public String decrypt(String keyName, String ciphertext) {
...
+	public List<Map<String, String>> decrypt(String keyName, List<String> ciphertexts) {
+
+		return decrypt(keyName, ciphertexts, null);
+	}
+	
+	@Override
+	public List<Map<String, String>> decrypt(String keyName, List<String> ciphertexts,
+			List<VaultTransitContext> transitRequests) {
+		
+		Boolean hasContext= false;
```
mp911de: See above.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultTransitTemplateIntegrationTests.java
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *	  http://www.apache.org/licenses/LICENSE-2.0
```
mp911de: No need to touch the license headers.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultTransitTemplateIntegrationTests.java
```diff
@@ -277,4 +282,126 @@ public void encryptAndRewrapShouldCreateCiphertextWithNonceAndContext() {
...
+		/**
+		 * check to see if decryption ended up with errors. this is special
+		 * handling when context is supplied. For the other use cases, the
+		 * handling should be done by the applications.
+		 */
+		if (decryptionContext != null) {
+			for (Map<String, String> entry : plaintextResponse) {
+
+				if (!StringUtils.isEmpty(entry.get("error"))) {
+					throw new VaultException(entry.get("error"));
```
mp911de: That shouldn't happen here. It's the responsibility of the Template API implementation to handle errors and throw exceptions.

mp911de:
Please include version guards (see [here](https://github.com/spring-projects/spring-vault/blob/aa8eb13f66e95a2e27c49ea41983bd6b9332a56b/spring-vault-core/src/test/java/org/springframework/vault/core/VaultTransitTemplateIntegrationTests.java#L140) for an example) to conditionally run the batch tests.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
pravsingh: makes lot of sense. will work on it.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
pravsingh: great suggestion. will take care.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultTransitTemplateIntegrationTests.java
pravsingh: originally I had this code inside the impl code. will put it back.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
pravsingh: How about Value Objects being like:

```java
@Data
class EncryptionRequest {
private String plaintext;
private VaultTransitContext context;
}
```

```java
@Data
class DecryptionRequest {
private String ciphertext;
private VaultTransitContext context;
}
```

A version of below APIs could be created to leverage above VOs:

```java
public String encrypt(String keyName, byte[] plaintext,
			VaultTransitContext transitRequest);
//newer version
public String encrypt(EncryptionRequest encryptionRequest);

```

```java
	public byte[] decrypt(String keyName, String ciphertext,
			VaultTransitContext transitRequest);

//newer version
public byte[] decrypt(DecryptionRequest decryptionRequest);

```

spring-vault-core/src/test/java/org/springframework/vault/core/VaultTransitTemplateIntegrationTests.java
mp911de: In the light of a `EncryptionResult`/`DecryptionResult` we could align the design of the value objects to a `Future`. Result types can expose a `get()` method that either returns the result or throws an exception. We should make sure to provide introspection methods like `isSuccess()` and `getCause()` for a smooth usage.

Does this make sense?

pravsingh: makes sense. very good suggestion.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
mp911de: `EncryptionRequest` should carry a binary payload. `byte[]` or `ByteBuffer` should be fine. The types should be immutable and expose factory methods like `of(String)` and `of(byte[])`. The context can be associated via a `with(VaultTransitContext)` creating a new object. I think a builder would not be a good fit here.

I'm also a bit concerned regarding the name. Usually, all `…Request` objects encapsulate the whole request. For batching, we don't want to include the key in the request because that would be confusing which key to use.

Please don't change existing API because it would break backward compatibility.

pravsingh: great suggestions on the factory pattern.
yes I wasn't suggesting to change existing APIs but was suggesting to create parallel APIs with new signature but eventually doing the same. I wouldn't do it as its not in batch api scope.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultDecryptionPayload.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Decryption Value Object used for encrypt() operations.
+ * 
+ * @author Praveendra Singh
+ *
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+public class VaultDecryptionPayload {
```
pravsingh: @mp911de 
I couldn't find better name and hence selected Payload instead of Request.
Please let me know if you would like to have different naming for the consistency purposes.

pravsingh:
@mp911de tried to incorporate changes you requested.
Please let me know if something got missed out in my understanding.
thanks

mp911de:
Thanks for the update. The change looks pretty cool now. I would suggest to change results of `VaultDecryptionResult decrypt(…)`  and `VaultEncryptionResult encrypt(…)` to `List<VaultDecryptionResult> decrypt(…)`  and `List<VaultEncryptionResult> encrypt(…)` to reduce responsibility to element-wise results.

Regarding Lombok usage, I'd like to keep public API parts without using actually it. In public API parts we just use `@EqualsAndHashCode` and `@ToString` – methods that do not harm if the code isn't visible. We use `@Data` only in private parts (such as `RawTransitKeyImpl` in `VaultTransitTemplate`).

pravsingh:
about the signature change recommendation.
VaultDecryptionResult holds the VaultResponse.
vaultResponse.getData().get("batch_results") holds the result of individual decryption responses.

Per your suggestion, VaultDecryptionResult will now contain individual responses instead of whole batch result.

will work on it.

pravsingh:
@mp911de 
updated code per your suggestions.
please review.

mp911de:
Thanks a lot. I will take the pull request from here. I think I'll do the merge it during the next week.

pravsingh:
awesome. thanks for your guidance Mark.

mp911de:
That's merged and polished now. Thanks for your contribution.

pravsingh:
thank you.

PR #162, opened 2017-10-15
Extend Transit API to support Hashing and Signing Operations

### LINKED ISSUE #148: Extend Transit API to Support Hashing and Signing Operations
Currently Spring Vault has support only for encryption/decryption and management of keys.
It would be nice to have hashing and signing methods in VaultTransitOperations.

Vault transit API is described at https://www.vaultproject.io/api/secret/transit/index.html

luander:
Transit API now has support to generate HMAC digests, sign and verify signed plaintext.

Closes #148

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
@@ -222,4 +222,67 @@ public interface VaultTransitOperations {
 	 * @see #rotate(String)
 	 */
 	String rewrap(String keyName, String ciphertext, VaultTransitContext transitContext);
+
+	/**
+	 * Generate HMAC digest of given data.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @return the digest of given data using sha2-256 hash algorithm and the named key.
+	 */
+	String generateHmac(String keyName, String plainText);
+
+	/**
+	 * Generate HMAC digest of given data.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param algorithm Specify the algorithm to be used for signing. Supported algorithms are:
+	 *                  sha2-224, sha2-256, sha2-384, sha2-512
+	 * @return the digest of given data using the specified hash hash algorithm and the named key.
+	 */
+	String generateHmac(String keyName, String plainText, String algorithm);
+
+
+	/**
+	 * Sign a String using a key from the vault using the SHA-256 algorithm.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @return Signature of the payload
+	 */
+	String sign(String keyName, String plainText);
+
+	/**
+	 * Sign a String using a key from the vault.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param algorithm Specify the algorithm to be used for signing. Supported algorithms are:
+	 *                  sha2-224, sha2-256, sha2-384, sha2-512
+	 * @return Signature of the payload
+	 */
+	String sign(String keyName, String plainText, String algorithm);
+
+	/**
+	 * Verify the validity of a signature in the vault.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param signature Signature to be verified
+	 * @return true if the signature is valid, false otherwise
+	 */
+	boolean verify(String keyName, String plainText, String signature);
+
+	/**
+	 * Verify the validity of a signature in the vault.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param signature Signature to be verified
+	 * @param algorithm Specify the algorithm to be used for signature verification.
+	 *                  Supported algorithms are: sha2-224, sha2-256, sha2-384, sha2-512
+	 * @return true if the signature is valid, false otherwise
+	 */
+	boolean verify(String keyName, String plainText, String signature, String algorithm);
 }
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
@@ -19,11 +19,16 @@ import java.util.List;
 
 import org.springframework.lang.Nullable;
 import org.springframework.vault.support.Ciphertext;
+import org.springframework.vault.support.Hmac;
 import org.springframework.vault.support.Plaintext;
 import org.springframework.vault.support.RawTransitKey;
+import org.springframework.vault.support.Signature;
 import org.springframework.vault.support.TransitKeyType;
 import org.springframework.vault.support.VaultDecryptionResult;
 import org.springframework.vault.support.VaultEncryptionResult;
+import org.springframework.vault.support.VaultHmacRequest;
+import org.springframework.vault.support.VaultSignRequest;
+import org.springframework.vault.support.VaultSignatureVerificationRequest;
 import org.springframework.vault.support.VaultTransitContext;
 import org.springframework.vault.support.VaultTransitKey;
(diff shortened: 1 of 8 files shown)


spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -222,4 +222,67 @@
...
+	/**
+	 * Generate HMAC digest of given data.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param algorithm Specify the algorithm to be used for signing. Supported algorithms are:
+	 *                  sha2-224, sha2-256, sha2-384, sha2-512
+	 * @return the digest of given data using the specified hash hash algorithm and the named key.
+	 */
+	String generateHmac(String keyName, String plainText, String algorithm);
```
mp911de: It would make sense to provide a `VaultHmacRequest` object that encapsulates `PlainText`, `algorithm` and possibly other options that apply to the HMAC operation.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -222,4 +222,67 @@
...
+	/**
+	 * Sign a String using a key from the vault.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param algorithm Specify the algorithm to be used for signing. Supported algorithms are:
+	 *                  sha2-224, sha2-256, sha2-384, sha2-512
+	 * @return Signature of the payload
+	 */
+	String sign(String keyName, String plainText, String algorithm);
```
mp911de: Same as above.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -222,4 +222,67 @@
...
+	 * Verify the validity of a signature in the vault.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param signature Signature to be verified
+	 * @param algorithm Specify the algorithm to be used for signature verification.
+	 *                  Supported algorithms are: sha2-224, sha2-256, sha2-384, sha2-512
+	 * @return true if the signature is valid, false otherwise
+	 */
+	boolean verify(String keyName, String plainText, String signature, String algorithm);
```
mp911de: We should introduce here a `VaultTransitVerificationRequest` type to encapsulate parameters and some sort of verification response. Returning `boolean` in the method above is perfectly fine if dealing with simple use-cases.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -222,4 +222,67 @@
...
+
+
+	/**
+	 * Sign a String using a key from the vault using the SHA-256 algorithm.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @return Signature of the payload
+	 */
+	String sign(String keyName, String plainText);
```
mp911de: `sign` should return a `Signature` result type to encapsulate the signature details. Similar to `Plantext` and `Ciphertext` types.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -222,4 +222,67 @@
...
+
+	/**
+	 * Verify the validity of a signature in the vault.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @param signature Signature to be verified
+	 * @return true if the signature is valid, false otherwise
+	 */
+	boolean verify(String keyName, String plainText, String signature);
```
mp911de: Let's accept here `Plaintext` and `Signature` types instead of plain strings.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
```diff
@@ -222,4 +222,67 @@
...
 	String rewrap(String keyName, String ciphertext, VaultTransitContext transitContext);
+
+	/**
+	 * Generate HMAC digest of given data.
+	 *
+	 * @param keyName must not be empty or {@literal null}.
+	 * @param plainText must not be empty or {@literal null}.
+	 * @return the digest of given data using sha2-256 hash algorithm and the named key.
+	 */
+	String generateHmac(String keyName, String plainText);
```
mp911de: Let's accept `Plaintext` as type here instead of a plain string.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -340,6 +343,61 @@ public String rewrap(String keyName, String ciphertext,
...
+	public String generateHmac(String keyName, String plainText, String algorithm) {
+		Assert.hasText(keyName, "KeyName must not be empty");
+		Assert.notNull(plainText, "Plain text must not be null");
+
+		Map<String, String> request = new LinkedHashMap<>();
+
+		request.put("input", Base64Utils.encodeToString(plainText.getBytes()));
+
+		return (String) vaultOperations.
+				write(String.format("%s/hmac/%s/%s", path, keyName, algorithm), request).getData()
```
mp911de: Let's not pass the algorithm as path variable but as part of the body. Key types like ed25519 have their own hashing algorithms and a path-based variant does not allow using the key-specifics.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -340,6 +343,61 @@ public String rewrap(String keyName, String ciphertext,
...
+
+	public String sign(String keyName, String plainText, String algorithm) {
+		Assert.hasText(keyName, "KeyName must not be empty");
+		Assert.notNull(plainText, "Plain text must not be null");
+
+		Map<String, String> request = new LinkedHashMap<>();
+
+		request.put("input", Base64Utils.encodeToString(plainText.getBytes()));
+
+		return (String) vaultOperations.
```
mp911de: See above.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
```diff
@@ -340,6 +343,61 @@ public String rewrap(String keyName, String ciphertext,
...
+		Assert.hasText(keyName, "KeyName must not be empty");
+		Assert.notNull(signature, "Signature must not be null");
+		Assert.notNull(plainText, "Input must not be null");
+
+		Map<String, String> request = new LinkedHashMap<>();
+
+		request.put("signature", signature);
+		request.put("input", Base64Utils.encodeToString(plainText.getBytes()));
+
+		return (boolean) vaultOperations.
```
mp911de: See above.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitOperations.java
luander: Hi Marc,
Thanks a lot for your feedback. Shall I overload the method with a Plaintext argument or change this one to only accept Plaintext?

mp911de: Let's just accept `Plaintext` instead of `String`. Having an overload is tempting but we would end up with three `generateHmac` methods. Same for the other new methods.

luander: Does it make sense here to return `Hmac` object as well?

mp911de: I hadn't a clear idea on this one yet. Considering the development, it would make sense to return a `Hmac` type. This would give us some consistency with the other methods.

mp911de:
That's awesome work. I'll take the pull-request from here.

luander:
Thanks Mark. The build failed for some versions of Vault though. Is that Ok?

mp911de:
I merged and polished your pull request. I added guards to the tests to run against the appropriate Vault version. Figuring out the right version in which features made it into Vault can be quite cumbersome.

PR #166, opened 2017-10-24
Implementation of Auth Backend with Kubernetes Service Account Token

everesio:
- KubeAuthentication integration tests are ignored so far, as local K8S setup (minikube) must be clarified


.travis.yml
```diff
@@ -33,9 +33,10 @@ before_install:
   - test ! -f ~/.m2/settings.xml || rm ~/.m2/settings.xml
 
 install:
+  - src/test/bash/minikube-ci-initialize.sh
```
mp911de: Let's stay consistent with underscores as delimiters.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubeServiceAccountTokenFile.java
```diff
@@ -0,0 +1,71 @@
...
+
+	public static final String DEFAULT_KUBERNETES_SERVICE_ACCOUNT_TOKEN_FILE = "/var/run/secrets/kubernetes.io/serviceaccount/token";
+
+	private final Resource resource;
+
+	public KubeServiceAccountTokenFile() {
+		this(DEFAULT_KUBERNETES_SERVICE_ACCOUNT_TOKEN_FILE);
+	}
+
+	public KubeServiceAccountTokenFile(final String fileName) {
```
mp911de: Nit: `final` isn't required here.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubeServiceAccountTokenFile.java
```diff
@@ -0,0 +1,71 @@
...
+		this(new FileSystemResource(fileName));
+	}
+
+	public KubeServiceAccountTokenFile(final File file) {
```
mp911de: Nit: `final` isn't required here.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubeServiceAccountTokenFile.java
```diff
@@ -0,0 +1,71 @@
...
+		this(new FileSystemResource(file));
+	}
+
+	public KubeServiceAccountTokenFile(final Resource resource) {
```
mp911de: Nit: `final` isn't required here.

src/test/bash/minikube-ci-initialize.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
mp911de: Let's stay with underscores instead of dashes for the file name.

mp911de:
I'm perfectly fine with using minikube on Travis, we should make sure to skip integration tests if Kubernetes is not running/available.

everesio:
I removed AuthenticationSteps and associated integration tests.

Normally the service accounts token are mounted automatically mounted into pod.  This behavior can be switched off by setting automountServiceAccountToken to false.

Every service account has an associated K8S secret. This secret holds the public CA of API server and the token.

everesio:
We could also move the build command to a separated script and retrieve the IP of the minikube.
This would be better than using hard coded 127.0.0.1.  

e.g.

.travis.yml
script: ./travis_build.sh


travis_build.sh:
MINIKUBE_IP=$(./minikube ip)
./mvnw clean verify -DMINIKUBE_IP=${MINIKUBE_IP} -P${PROFILE:-ci}

mp911de:
Would `./mvnw clean verify -DMINIKUBE_IP=$(./minikube ip) -P${PROFILE:-ci}` work? No need to add further indirections if not required. I don't really have an issue with `127.0.0.1`, with Travis, we can export the IP as env variable and propagate it to the init scripts.

everesio:
OK. I will rework my last change.

everesio:
I have just started implementing KUBERNETES binding in the spring-cloud-vault. Without the AuthenticationSteps (I removed earlier) I am getting an error (with AuthenticationSteps it works fine).
I am just a little confused ? 

`Running org.springframework.cloud.vault.config.VaultConfigWithContextTests
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.436 sec - in org.springframework.cloud.vault.config.VaultConfigWithContextTests
Running org.springframework.cloud.vault.config.VaultConfigKubernetesTests
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.269 sec <<< FAILURE! - in org.springframework.cloud.vault.config.VaultConfigKubernetesTests
contextLoads(org.springframework.cloud.vault.config.VaultConfigKubernetesTests)  Time elapsed: 0.003 sec  <<< ERROR!
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'reactiveVaultTemplate' defined in org.springframework.cloud.vault.config.ReactiveVaultBootstrapConfiguration: Unsatisfied dependency expressed through method 'reactiveVaultTemplate' parameter 0; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'vaultTokenSupplier' defined in org.springframework.cloud.vault.config.ReactiveVaultBootstrapConfiguration: Unsatisfied dependency expressed through method 'vaultTokenSupplier' parameter 0; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.vault.authentication.AuthenticationStepsFactory' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'vaultTokenSupplier' defined in org.springframework.cloud.vault.config.ReactiveVaultBootstrapConfiguration: Unsatisfied dependency expressed through method 'vaultTokenSupplier' parameter 0; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.vault.authentication.AuthenticationStepsFactory' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.vault.authentication.AuthenticationStepsFactory' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
`

mp911de:
`AuthenticationSteps` is (a `AuthenticationStepsFactory` bean) is only required if you enable the reactive configuration in Spring Cloud Vault (actually, this discussion would rather belong into https://github.com/spring-cloud/spring-cloud-vault/). Since we removed `AuthenticationSteps` from the Kubernetes auth, there is no `AuthenticationStepsFactory` available anymore.

I think we can make it work anyway, if we make the right assumptions. If the JWT associated with the service account never changes while a pod runs, we can pre-fetch the JWT during config bootstrap and provide it in a static way. 

That would require to create a variant of  `AuthenticationStepsFactory`  that takes the JWT token, something like:

```java
Map<String, String> map = … // build the Map with role and jwt
AuthenticationSteps.fromSupplier(() -> map).login("kubernetes/login")
```

everesio:
In such a case, does the implementation of AuthenticationStepsFactory belong to spring-cloud-vault or sprig-vault (reverting the removal of [createAuthenticationSteps](https://github.com/spring-projects/spring-vault/pull/166/commits/6c0aa0e68d6018c2541baf8ec6d463a9f1c6605d#diff-f3dfa3d0fd6fcba45a9b21112d77eff4R81)) ?

mp911de:
Belongs in to Spring Cloud Vault. I can take care of that part once Kubernetes support is in place.

Am 28. Okt. 2017, 12:23 +0200 schrieb Michal Budzyn <notifications@github.com>:

everesio:
OK. Are there any open points preventing merging ? 
BTW. As we still use spring boot 1.5, we would like to have Kubernetes Auth also in older release too. Are there obstacles preventing it ?

mp911de:
The following steps are that I need to get familiar with minikube and then I will go on with the merge. 

Spring Vault 1.1.x was the last feature release of the 1.x branch. All subsequent releases of 1.1.x are maintenance releases without adding any further features. Spring Cloud Vault 1.1.0 is not GA yet so we can add Kubernetes auth there (basically port your code to Spring Cloud Vault as non-public API) – which would work for Boot 1.5 but only if you use Spring Cloud Vault.

For Spring Cloud Vault 2.0 we can pick up the Kubernetes auth provided in Spring Vault.

everesio:
Great. I will do some work on Spring Cloud Vault 1.1.0

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubeServiceAccountTokenFile.java
```diff
@@ -0,0 +1,71 @@
...
+		this(new FileSystemResource(file));
+	}
+
+	public KubeServiceAccountTokenFile(Resource resource) {
+		this.resource = resource;
+	}
+
+	@Override
+	public String getKubeJwt() {
+		try (InputStream is = resource.getInputStream()) {
```
mp911de: Is it safe to assume the service account token never changes when a pod is running?

everesio: As far as I know token is not changed (see also [Service Account Key Rotation #20165](https://github.com/kubernetes/kubernetes/issues/20165)).

When you create a service account, a secret is created. This secret is mounted to the pod.   
In general secrets cannot be modified (only delete/recreate works).  

If you delete and recreate the serviceaccount a new secret will be created, but the old mount persist e.g. 

```
1. kubectl create serviceaccount hello
2. kubectl get serviceaccounts/hello -o yaml
kind: ServiceAccount
metadata:
  uid: 7b60565f-bf1f-11e7-b013-080027efc8f4
secrets:
- name: hello-token-rv80h

3. run hello-minikube with serviceaccount hello
4. kubectl get pods hello-minikube-1872458443-v4ld0 -o yaml

spec:
  containers:
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: hello-token-rv80h
      readOnly: true
  volumes:
  - name: hello-token-rv80h
    secret:
      defaultMode: 420
      secretName: hello-token-rv80h

5. kubectl delete serviceaccount hello && kubectl create serviceaccount hello
6. kubectl get serviceaccounts/hello -o yaml

kind: ServiceAccount
metadata:
  uid: 966fb602-bf25-11e7-b013-080027efc8f4
secrets:
- name: hello-token-h713z

7.  kubectl get pods hello-minikube-1872458443-v4ld0 -o yaml
kind: Pod
spec:
  containers:
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: hello-token-rv80h
      readOnly: true
  volumes:
  - name: hello-token-rv80h
    secret:
      defaultMode: 420
      secretName: hello-token-rv80h
```

mp911de: Thanks. That's good news in the light of `AuthenticationSteps`.

everesio: You are welcome. If you don't mind, I would copy to this PR the changes I made in [Spring Cloud V #176](https://github.com/spring-cloud/spring-cloud-vault/pull/176) in the local_run_k8s.sh and minikube_ci_initialize.sh ?

mp911de:
Thanks for your contribution. That's now merged and polished.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubeServiceAccountTokenFile.java
mp911de: Feel free to update the PR in Spring Cloud Vault. Continuing the discussion in https://github.com/spring-cloud/spring-cloud-vault/pull/176.

PR #216, opened 2018-03-12
Initial cut of revisions to Exception Hierarchy

mikebell90:
1. Virtually all exceptions (with a few exceptions, please review) were remapped to this hierarchy

VaultException (mostly serves as a root, few instantiated)
-- VaultClientException - where the state or a mapping error says "Hey this isn't Vault per se, but a bad state"
-- VaultRemoteException - mostly a container class but represents fatal results from a call to Vault
----- VaultHttpExceptions - Most Vault calls map here.

2. I agree a cleaner mapper is wise. Right now things are a mess! I agree with your concept, but felt it belonged in another PR. One of the issues is some things have an exception mapper (VaultResponses), others don't.

3. I couldn't run most tests, because apparently they require a local workdir and Vault -please clarify how I can run this. As an aside, while I don't have it working perfectly - testContainers.org is a lovely way to run integration tests with a dockerized Vault.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
@@ -228,7 +228,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 		}
 		catch (HttpStatusCodeException e) {
 			throw new VaultHttpException(String.format("Cannot login using AppRole: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+					VaultResponses.getError(e.getResponseBodyAsString())), e);
 		}
 	}
 
@@ -253,7 +253,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 			catch (HttpStatusCodeException e) {
 				throw new VaultHttpException(String.format(
 						"Cannot get Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+						VaultResponses.getError(e.getResponseBodyAsString())), e);
 			}
 		}
 
@@ -276,7 +276,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 			catch (HttpStatusCodeException e) {
 				throw new VaultHttpException(String.format(
 						"Cannot unwrap Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+						VaultResponses.getError(e.getResponseBodyAsString())), e);
 			}
 		}
 
@@ -302,7 +302,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 			catch (HttpStatusCodeException e) {
 				throw new VaultHttpException(String.format(
 						"Cannot get Secret id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+						VaultResponses.getError(e.getResponseBodyAsString())), e);
 			}
 		}
 
@@ -325,7 +325,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 			catch (HttpStatusCodeException e) {
 				throw new VaultHttpException(String.format(
 						"Cannot unwrap Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+						VaultResponses.getError(e.getResponseBodyAsString())), e);
 			}
 		}
 
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
@@ -15,12 +15,14 @@
  */
 package org.springframework.vault.authentication;
 
+import static org.springframework.vault.authentication.AuthenticationSteps.HttpRequestBuilder.get;
+import static org.springframework.vault.authentication.AuthenticationSteps.HttpRequestBuilder.post;
+
 import java.util.HashMap;
 import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
@@ -37,13 +39,12 @@ import org.springframework.vault.authentication.AppRoleTokens.Pull;
 import org.springframework.vault.authentication.AppRoleTokens.Wrapped;
 import org.springframework.vault.authentication.AuthenticationSteps.Node;
 import org.springframework.vault.client.VaultResponses;
+import org.springframework.vault.exceptions.VaultClientException;
+import org.springframework.vault.exceptions.VaultHttpException;
 import org.springframework.vault.support.VaultResponse;
 import org.springframework.vault.support.VaultToken;
 import org.springframework.web.client.HttpStatusCodeException;
 import org.springframework.web.client.RestOperations;
-
-import static org.springframework.vault.authentication.AuthenticationSteps.HttpRequestBuilder.get;
-import static org.springframework.vault.authentication.AuthenticationSteps.HttpRequestBuilder.post;
 
 /**
  * AppRole implementation of {@link ClientAuthentication}. RoleId and SecretId (optional)
@@ -105,7 +106,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 		if ((roleId instanceof Wrapped || roleId instanceof Pull)
 				&& (secretId instanceof Wrapped || secretId instanceof Pull)) {
 
-			throw new IllegalArgumentException(
+			throw new VaultClientException(
 					"RoleId and SecretId are both configured to obtain their values from initial Vault request. AuthenticationSteps supports currently only fetching of a single element.");
 		}
 
@@ -181,7 +182,7 @@ public class AppRoleAuthentication implements ClientAuthentication,
 			});
 		}
 
-		throw new IllegalArgumentException(String.format(
+		throw new VaultClientException(String.format(
 				"Provided RoleId/SecretId setup not supported. RoleId: %s, SecretId: %s",
 				roleId, secretId));
 	}
@@ -226,8 +227,8 @@ public class AppRoleAuthentication implements ClientAuthentication,
 			return LoginTokenUtil.from(response.getAuth());
 		}
 		catch (HttpStatusCodeException e) {
-			throw new VaultException(String.format("Cannot login using AppRole: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())));
+			throw new VaultHttpException(String.format("Cannot login using AppRole: %s",
+					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 		}
 	}
 
@@ -250,9 +251,9 @@ public class AppRoleAuthentication implements ClientAuthentication,
 				return (String) entity.getBody().getRequiredData().get("role_id");
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
 						"Cannot get Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 			}
 		}
 
@@ -273,13 +274,13 @@ public class AppRoleAuthentication implements ClientAuthentication,
 				return (String) response.getRequiredData().get("role_id");
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
 						"Cannot unwrap Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 			}
 		}
 
-		throw new IllegalArgumentException("Unknown RoleId configuration: " + roleId);
+		throw new VaultClientException("Unknown RoleId configuration: " + roleId);
 	}
 
 	private String getSecretId(SecretId secretId) {
@@ -299,9 +300,9 @@ public class AppRoleAuthentication implements ClientAuthentication,
 				return (String) response.getRequiredData().get("secret_id");
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
 						"Cannot get Secret id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 			}
 		}
 
@@ -322,13 +323,13 @@ public class AppRoleAuthentication implements ClientAuthentication,
 				return (String) response.getRequiredData().get("secret_id");
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
 						"Cannot unwrap Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 			}
 		}
 
-		throw new IllegalArgumentException("Unknown SecretId configuration: " + secretId);
+		throw new VaultClientException("Unknown SecretId configuration: " + secretId);
 	}
 
 	private static HttpHeaders createHttpHeaders(VaultToken token) {
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
@@ -108,7 +108,7 @@ public class AuthenticationStepsExecutor implements ClientAuthentication {
 				throw new VaultHttpException(String.format(
 						"HTTP request %s in state %s failed with Status %s and body %s",
 						o, state, e.getStatusCode(),
-						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+						VaultResponses.getError(e.getResponseBodyAsString())), e);
 			}
 			catch (VaultException e) {
 				throw e;
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
@@ -30,6 +30,7 @@ import org.springframework.vault.authentication.AuthenticationSteps.Node;
 import org.springframework.vault.authentication.AuthenticationSteps.OnNextStep;
 import org.springframework.vault.authentication.AuthenticationSteps.SupplierStep;
 import org.springframework.vault.client.VaultResponses;
+import org.springframework.vault.exceptions.VaultHttpException;
 import org.springframework.vault.support.VaultResponse;
 import org.springframework.vault.support.VaultToken;
 import org.springframework.web.client.HttpStatusCodeException;
@@ -104,10 +105,13 @@ public class AuthenticationStepsExecutor implements ClientAuthentication {
 				}
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
 						"HTTP request %s in state %s failed with Status %s and body %s",
 						o, state, e.getStatusCode(),
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+			}
+			catch (VaultException e) {
+				throw e;
 			}
 			catch (RuntimeException e) {
 				throw new VaultException(String.format(
@@ -126,7 +130,7 @@ public class AuthenticationStepsExecutor implements ClientAuthentication {
 			return LoginTokenUtil.from(response.getAuth());
 		}
 
-		throw new IllegalStateException(String.format(
+		throw new VaultException(String.format(
 				"Cannot retrieve VaultToken from authentication chain. Got instead %s",
 				state));
 	}
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsOperator.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsOperator.java
@@ -31,6 +31,8 @@ import org.springframework.vault.authentication.AuthenticationSteps.MapStep;
 import org.springframework.vault.authentication.AuthenticationSteps.Node;
 import org.springframework.vault.authentication.AuthenticationSteps.OnNextStep;
 import org.springframework.vault.authentication.AuthenticationSteps.SupplierStep;
+import org.springframework.vault.exceptions.VaultClientException;
+import org.springframework.vault.exceptions.VaultRemoteException;
 import org.springframework.vault.support.VaultResponse;
 import org.springframework.vault.support.VaultToken;
 import org.springframework.web.reactive.function.client.WebClient;
@@ -127,13 +129,13 @@ public class AuthenticationStepsOperator implements VaultTokenSupplier {
 						return LoginTokenUtil.from(response.getAuth());
 					}
 
-					throw new IllegalStateException(
+					throw new VaultException(
 							String.format(
 									"Cannot retrieve VaultToken from authentication chain. Got instead %s",
 									stateObject));
 				})
 				.onErrorMap(
-						t -> new VaultException(
+						t -> t instanceof VaultException ? t : new VaultException(
 								"Cannot retrieve VaultToken from authentication chain", t));
 	}
 
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
@@ -187,7 +187,7 @@ public class AwsEc2Authentication implements ClientAuthentication,
 		}
 		catch (HttpStatusCodeException e) {
 			throw new VaultHttpException(String.format("Cannot login using AWS-EC2: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+					VaultResponses.getError(e.getResponseBodyAsString())), e);
 		}
 	}
 
@@ -217,7 +217,7 @@ public class AwsEc2Authentication implements ClientAuthentication,
 		catch (HttpStatusCodeException e) {
 			throw new VaultHttpException(String.format(
 					"Cannot obtain Identity Document from %s",
-					options.getIdentityDocumentUri()), e, e.getStatusCode());
+					options.getIdentityDocumentUri()), e);
 		}
 		catch (RestClientException e) {
 			throw new VaultRemoteException(String.format(
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
@@ -29,6 +29,8 @@ import org.springframework.util.StringUtils;
 import org.springframework.vault.VaultException;
 import org.springframework.vault.authentication.AuthenticationSteps.HttpRequestBuilder;
 import org.springframework.vault.client.VaultResponses;
+import org.springframework.vault.exceptions.VaultHttpException;
+import org.springframework.vault.exceptions.VaultRemoteException;
 import org.springframework.vault.support.VaultResponse;
 import org.springframework.vault.support.VaultToken;
 import org.springframework.web.client.HttpStatusCodeException;
@@ -184,8 +186,8 @@ public class AwsEc2Authentication implements ClientAuthentication,
 			return LoginTokenUtil.from(response.getAuth());
 		}
 		catch (HttpStatusCodeException e) {
-			throw new VaultException(String.format("Cannot login using AWS-EC2: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())));
+			throw new VaultHttpException(String.format("Cannot login using AWS-EC2: %s",
+					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 		}
 	}
 
@@ -212,8 +214,13 @@ public class AwsEc2Authentication implements ClientAuthentication,
 
 			return login;
 		}
+		catch (HttpStatusCodeException e) {
+			throw new VaultHttpException(String.format(
+					"Cannot obtain Identity Document from %s",
+					options.getIdentityDocumentUri()), e, e.getStatusCode());
+		}
 		catch (RestClientException e) {
-			throw new VaultException(String.format(
+			throw new VaultRemoteException(String.format(
 					"Cannot obtain Identity Document from %s",
 					options.getIdentityDocumentUri()), e);
 		}
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthentication.java
@@ -138,7 +138,7 @@ public class AwsIamAuthentication implements ClientAuthentication {
 		}
 		catch (HttpStatusCodeException e) {
 			throw new VaultHttpException(String.format("Cannot login using AWS-IAM: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+					VaultResponses.getError(e.getResponseBodyAsString())), e);
 		}
 	}
 
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthentication.java
@@ -37,6 +37,8 @@ import org.springframework.util.Base64Utils;
 import org.springframework.util.StringUtils;
 import org.springframework.vault.VaultException;
 import org.springframework.vault.client.VaultResponses;
+import org.springframework.vault.exceptions.VaultClientException;
+import org.springframework.vault.exceptions.VaultHttpException;
 import org.springframework.vault.support.VaultResponse;
 import org.springframework.vault.support.VaultToken;
 import org.springframework.web.client.HttpStatusCodeException;
@@ -135,8 +137,8 @@ public class AwsIamAuthentication implements ClientAuthentication {
 			return LoginTokenUtil.from(response.getAuth());
 		}
 		catch (HttpStatusCodeException e) {
-			throw new VaultException(String.format("Cannot login using AWS-IAM: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())));
+			throw new VaultHttpException(String.format("Cannot login using AWS-IAM: %s",
+					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 		}
 	}
 
@@ -194,7 +196,7 @@ public class AwsIamAuthentication implements ClientAuthentication {
 			return OBJECT_MAPPER.writeValueAsString(map);
 		}
 		catch (JsonProcessingException e) {
-			throw new IllegalStateException("Cannot serialize headers to JSON", e);
+			throw new VaultClientException("Cannot serialize headers to JSON", e);
 		}
 	}
 
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/IpAddressUserId.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/IpAddressUserId.java
@@ -19,6 +19,8 @@ package org.springframework.vault.authentication;
 import java.io.IOException;
 import java.net.InetAddress;
 
+import org.springframework.vault.exceptions.VaultClientException;
+
 /**
  * Mechanism to generate a SHA-256 hashed and hex-encoded representation of the IP
  * address. Can be calculated with {@code echo -n 192.168.99.1 | sha256sum}.
@@ -34,7 +36,7 @@ public class IpAddressUserId implements AppIdUserIdMechanism {
 			return Sha256.toSha256(InetAddress.getLocalHost().getHostAddress());
 		}
 		catch (IOException e) {
-			throw new IllegalStateException(e);
+			throw new VaultClientException("While creating userId", e);
 		}
 	}
 }
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
@@ -107,7 +107,7 @@ public class KubernetesAuthentication implements ClientAuthentication,
 		}
 		catch (HttpStatusCodeException e) {
 			throw new VaultHttpException(String.format("Cannot login using kubernetes: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
+					VaultResponses.getError(e.getResponseBodyAsString())), e);
 		}
 	}
 
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
@@ -24,6 +24,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.util.Assert;
 import org.springframework.vault.VaultException;
 import org.springframework.vault.client.VaultResponses;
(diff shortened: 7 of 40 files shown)


mikebell90:
I did find BTW Mark, that all of this hierarchy was needed. I also put my new exceptions in a "exceptions" package. Unfortunately we can't move VaultException or VaultTokenLookupException there without breaking compatibility.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
```diff
@@ -184,8 +186,8 @@ private VaultToken createTokenUsingAwsEc2() {
 			return LoginTokenUtil.from(response.getAuth());
 		}
 		catch (HttpStatusCodeException e) {
-			throw new VaultException(String.format("Cannot login using AWS-EC2: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())));
+			throw new VaultHttpException(String.format("Cannot login using AWS-EC2: %s",
+					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
```
mp911de: In some sense, `VaultHttpException: Cannot obtain Identity Document from` is the cause for a `VaultLoginException`

spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthentication.java
```diff
@@ -194,7 +196,7 @@ private static String getSignedHeaders(AwsIamAuthenticationOptions options) {
 			return OBJECT_MAPPER.writeValueAsString(map);
 		}
 		catch (JsonProcessingException e) {
-			throw new IllegalStateException("Cannot serialize headers to JSON", e);
+			throw new VaultClientException("Cannot serialize headers to JSON", e);
```
mp911de: 👍

spring-vault-core/src/main/java/org/springframework/vault/authentication/IpAddressUserId.java
```diff
@@ -34,7 +36,7 @@ public String createUserId() {
 			return Sha256.toSha256(InetAddress.getLocalHost().getHostAddress());
 		}
 		catch (IOException e) {
-			throw new IllegalStateException(e);
+			throw new VaultClientException("While creating userId", e);
```
mp911de: This could be a `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
```diff
@@ -105,8 +106,8 @@ public VaultToken login() throws VaultException {
 			return LoginTokenUtil.from(response.getAuth());
 		}
 		catch (HttpStatusCodeException e) {
-			throw new VaultException(String.format("Cannot login using kubernetes: %s",
-					VaultResponses.getError(e.getResponseBodyAsString())));
+			throw new VaultHttpException(String.format("Cannot login using kubernetes: %s",
```
mp911de: See above.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesServiceAccountTokenFile.java
```diff
@@ -96,7 +97,7 @@ public KubernetesServiceAccountTokenFile(Resource resource) {
 			this.token = readToken(resource);
 		}
 		catch (IOException e) {
-			throw new VaultException(String.format(
+			throw new VaultClientException(String.format(
```
mp911de: `VaultException` should be fine here.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
```diff
@@ -221,10 +223,10 @@ protected boolean renewToken() {
 				return false;
 			}
 
-			throw new VaultException(VaultResponses.getError(e.getResponseBodyAsString()));
+			throw new VaultHttpException(VaultResponses.getError(e.getResponseBodyAsString()), e.getStatusCode());
```
mp911de: Not exactly sure, but that could be a place for a more specific exception like a `VaultSessionException`. Same applies for the code below. Retaining the original exception would be a good overall improvement.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LoginTokenAdapter.java
```diff
@@ -104,7 +104,7 @@ static LoginToken augmentWithSelfLookup(RestOperations restOperations,
 		catch (HttpStatusCodeException e) {
 			throw new VaultTokenLookupException(String.format(
 					"Token self-lookup failed: %s %s", e.getStatusCode(),
-					VaultResponses.getError(e.getResponseBodyAsString())));
+					VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
```
mp911de: This change assumes that lookup is caused by a HTTP status code. Looking at the code above, `VaultTokenLookupException` applies, too with the slight difference that HTTP status was 200 or something other from the 2xx series. We should rather remove the HTTP assumption from `VaultTokenLookupException` and retain the original cause.

spring-vault-core/src/main/java/org/springframework/vault/authentication/MacAddressUserId.java
```diff
@@ -128,7 +129,7 @@ public String createUserId() {
 							"Cannot determine NetworkInterface"));
 		}
 		catch (IOException e) {
-			throw new IllegalStateException(e);
+			throw new VaultClientException("While creating userId", e);
```
mp911de: Rather `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/MacAddressUserId.java
```diff
@@ -172,7 +173,7 @@ private static boolean matchesHint(String hint, NetworkInterface networkInterfac
 			return Optional.ofNullable(it.getHardwareAddress());
 		}
 		catch (SocketException e) {
-			throw new IllegalStateException(String
+			throw new VaultClientException(String
```
mp911de: `VaultException` here, too.

spring-vault-core/src/main/java/org/springframework/vault/authentication/ReactiveLifecycleAwareSessionManager.java
```diff
@@ -409,7 +411,7 @@ private OneShotTrigger createTrigger(VaultToken token) {
 						e -> {
 							return new VaultTokenLookupException(String.format(
 									"Token self-lookup failed: %s %s", e.getStatusCode(),
-									VaultResponses.getError(e.getResponseBodyAsString())));
+									VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
```
mp911de: See above.

spring-vault-core/src/main/java/org/springframework/vault/authentication/VaultTokenLookupException.java
```diff
@@ -15,22 +15,23 @@
...
+import org.springframework.vault.exceptions.VaultHttpException;
 
 /**
  * Exception thrown if a token self-lookup fails via {@code auth/token/lookup-self}.
  *
  * @author Mark Paluch
  * @since 2.0
  */
-public class VaultTokenLookupException extends VaultException {
+public class VaultTokenLookupException extends VaultHttpException {
```
mp911de: We should keep extend from `VaultException`. Changing to `VaultHttpException` makes the assumption that lookup is always related to a `HttpStatus` which doesn't apply in every case.

spring-vault-core/src/main/java/org/springframework/vault/client/VaultEndpoint.java
```diff
@@ -92,7 +93,7 @@ public static VaultEndpoint from(URI uri) {
 					: uri.getPort());
 		}
 		catch (MalformedURLException e) {
-			throw new IllegalArgumentException(String.format(
```
mp911de: This should stay IAE.

spring-vault-core/src/main/java/org/springframework/vault/client/VaultResponses.java
```diff
@@ -188,7 +190,7 @@ public HttpHeaders getHeaders() {
 			});
 		}
 		catch (IOException e) {
-			throw new IllegalStateException(e);
+			throw new VaultClientException("While mapping object", e);
```
mp911de: How about `VaultClientException("Cannot unwrap response", e);`?

spring-vault-core/src/main/java/org/springframework/vault/config/AbstractReactiveVaultConfiguration.java
```diff
@@ -137,7 +138,7 @@ protected VaultTokenSupplier vaultTokenSupplier() {
 			return CachingVaultTokenSupplier.of(stepsOperator);
 		}
 
-		throw new IllegalStateException(
+		throw new VaultClientException(
```
mp911de: This should stay `IllegalStateException`

spring-vault-core/src/main/java/org/springframework/vault/config/ClientHttpConnectorFactory.java
```diff
@@ -87,7 +88,7 @@ private static void configureSsl(SslConfiguration sslConfiguration,
 			}
 		}
 		catch (GeneralSecurityException | IOException e) {
-			throw new IllegalStateException(e);
+			throw new VaultClientException("While configuring ssl", e);
```
mp911de: Let's keep it as `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/config/ClientHttpRequestFactoryFactory.java
```diff
@@ -108,11 +109,8 @@ public static ClientHttpRequestFactory create(ClientOptions options,
 				return Netty.usingNetty(options, sslConfiguration);
 			}
 		}
-		catch (GeneralSecurityException e) {
-			throw new IllegalStateException(e);
-		}
-		catch (IOException e) {
-			throw new IllegalStateException(e);
+		catch (GeneralSecurityException | IOException e) {
+			throw new VaultClientException("While configuring ClientRequestFactory", e);
```
mp911de: How about `new VaultClientException("Cannot create ClientRequestFactory", e)`?

spring-vault-core/src/main/java/org/springframework/vault/config/EnvironmentVaultConfiguration.java
```diff
@@ -174,7 +175,7 @@ public VaultEndpoint vaultEndpoint() {
 			return VaultEndpoint.from(URI.create(uri));
 		}
 
-		throw new IllegalStateException("Vault URI (vault.uri) is null");
+		throw new VaultClientException("Vault URI (vault.uri) is null");
```
mp911de: Let's stick with `IllegalStateException`.

spring-vault-core/src/main/java/org/springframework/vault/config/EnvironmentVaultConfiguration.java
```diff
@@ -233,7 +234,7 @@ public ClientAuthentication clientAuthentication() {
 		case KUBERNETES:
 			return kubeAuthentication();
 		default:
-			throw new IllegalStateException(String.format(
+			throw new VaultClientException(String.format(
```
mp911de: Let's stick with `IllegalStateException`.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultSysTemplate.java
```diff
@@ -228,7 +230,7 @@ public Policy getPolicy(String name) throws VaultException {
 					return null;
 				}
 
-				throw e;
+				throw new VaultHttpException(String.format("While getting policy %s", name), e, e.getStatusCode());
```
mp911de: Exception translation happens already in `VaultTemplate.doWithSession(…)`

spring-vault-core/src/main/java/org/springframework/vault/exceptions/VaultHttpException.java
```diff
@@ -0,0 +1,55 @@
...
+@SuppressWarnings("serial")
+public class VaultHttpException extends VaultRemoteException {
+	private final HttpStatus httpStatus;
+
+	/**
+	 * Create a {@code VaultException} with the specified detail message.
+	 *
+	 * @param msg the detail message.
+	 */
+	public VaultHttpException(String msg, HttpStatus httpStatus) {
```
mp911de: We're duplicating functionality in some sense. How about just retaining the cause (which is a HttpStatusException) and omitting `HttpStatus` fields/getters? Users interested in introspection can interact with the cause.

spring-vault-core/src/main/java/org/springframework/vault/exceptions/package-info.java
```diff
@@ -0,0 +1,7 @@
+/**
+ * Spring Vault Client Exceptions. Some Exceptions still remain in other packages
+ * for compatibility purposes.
+ */
+@org.springframework.lang.NonNullApi
+@org.springframework.lang.NonNullFields
+package org.springframework.vault.exceptions;
```
mp911de: We should put exceptions into the packages where they are exported/primarily thrown. An exception package is always a dependency magnet that does not properly reflect the structure of the code.

spring-vault-core/src/main/java/org/springframework/vault/repository/query/VaultQueryCreator.java
```diff
@@ -162,7 +163,7 @@ protected VaultQuery and(Part part, VaultQuery base, Iterator<Object> parameters
 			return new Criteria<>(accessor.nextString(parameters),
 					(value, it) -> !it.equals(value));
 		default:
-			throw new IllegalArgumentException("Unsupported keyword!");
+			throw new VaultClientException("Unsupported keyword!");
```
mp911de: IAE is the correct one here.

spring-vault-core/src/main/java/org/springframework/vault/support/AbstractResult.java
```diff
@@ -85,8 +85,8 @@ public V get() {
 		if (isSuccessful()) {
 			return get0();
 		}
-
-		throw new VaultException(exception.getMessage());
+		// Mark: This seemed wrong, and lost information
```
mp911de: The proper implementation would be throwing a completion exception similar to a `Future`. Let's keep this one out of the pull request and address the issue in a separate ticket.

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
```diff
@@ -106,7 +107,7 @@ public X509Certificate getX509Certificate() {
 			return KeystoreUtil.getCertificate(bytes);
 		}
 		catch (IOException | CertificateException e) {
-			throw new VaultException("Cannot create Certificate from certificate", e);
```
mp911de: Let's stick with `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
```diff
@@ -123,7 +124,7 @@ public X509Certificate getX509IssuerCertificate() {
 			return KeystoreUtil.getCertificate(bytes);
 		}
 		catch (IOException | CertificateException e) {
-			throw new VaultException(
```
mp911de: Let's stick with `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
```diff
@@ -141,7 +142,7 @@ public KeyStore createTrustStore() {
 					getX509IssuerCertificate());
 		}
 		catch (GeneralSecurityException | IOException e) {
-			throw new VaultException("Cannot create KeyStore", e);
```
mp911de: Let's stick with `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
```diff
@@ -93,7 +94,7 @@ public KeySpec getPrivateKeySpec() {
 			return KeystoreUtil.getRSAKeySpec(bytes);
 		}
 		catch (IOException e) {
-			throw new VaultException("Cannot create KeySpec from private key", e);
+			throw new VaultClientException("Cannot create KeySpec from private key", e);
```
mp911de: Let's stick with `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
```diff
@@ -114,7 +115,7 @@ public KeyStore createKeyStore(String keyAlias) {
 					getX509Certificate(), getX509IssuerCertificate());
 		}
 		catch (GeneralSecurityException | IOException e) {
-			throw new VaultException("Cannot create KeyStore", e);
+			throw new VaultClientException("Cannot create KeyStore", e);
```
mp911de: Let's stick with `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/support/KeystoreUtil.java
```diff
@@ -32,6 +32,8 @@
 import java.util.Collections;
 import java.util.List;
 
+import org.springframework.vault.exceptions.VaultClientException;
```
mp911de: ISE is the appropriate exception in all cases in `KeyStoreUtil`

spring-vault-core/src/main/java/org/springframework/vault/support/Policy.java
```diff
@@ -51,6 +51,7 @@
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
+import org.springframework.vault.exceptions.VaultClientException;
```
mp911de: IAE is the correct exception for all changed cases in here.

spring-vault-core/src/main/java/org/springframework/vault/support/SslConfiguration.java
```diff
@@ -455,7 +456,7 @@ public String getDescription() {
 
 		@Override
 		public InputStream getInputStream() throws IOException {
-			throw new UnsupportedOperationException("Empty resource");
+			throw new VaultClientException("Empty resource");
```
mp911de: `UnsupportedOperationException` is the correct exception here.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultResponseSupport.java
```diff
@@ -81,7 +82,7 @@
 			return auth;
 		}
 
-		throw new IllegalStateException("Auth field is empty");
+		throw new VaultClientException("Auth field is empty");
```
mp911de: ISE is the correct exception here.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultResponseSupport.java
```diff
@@ -109,7 +110,7 @@ public T getRequiredData() {
 			return data;
 		}
 
-		throw new IllegalStateException("Data field is empty");
+		throw new VaultClientException("Data field is empty");
```
mp911de: ISE is the correct exception here.

spring-vault-core/src/test/java/org/springframework/vault/util/Settings.java
```diff
@@ -69,7 +70,7 @@ public static File findWorkDir(File directory) {
 			searchLevel = searchLevel.getParentFile();
 		}
 
-		throw new IllegalStateException(String.format(
+		throw new VaultClientException(String.format(
```
mp911de: ISE is the correct exception here.

mikebell90:
Overall response (before reading your specifics, just your general):

Totally agreed that reasonable people can come up with different solutions. I deviated in one direction, because it seemed like at least half of the illegalxxxoperations exist BECAUSE there wasn't a hierarchy. 

But this is certainly where reasonable people can back and forth and come to consensus, so now I'll look at your specific comments

spring-vault-core/src/main/java/org/springframework/vault/exceptions/VaultClientException.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * An exception that is Spring Vault Client specific, but not related to
+ * a remote call to Vault. Common causes include message conversion failures,
+ * incorrect/impossible parameters.
+ *
+ * @author Michael Bell
+ */
+@SuppressWarnings("serial")
+public class VaultClientException extends VaultException {
```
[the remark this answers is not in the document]
mikebell90: The idea is that normally any known local reason for failure, such as pojo converted BEFORE request, or some missing parameter is VaultClientException. VaultException is simply a container root. If one drops this, one drops the ability to make this differentiation. I think it's a mistake, but continuing on...

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -105,7 +106,7 @@ public static AuthenticationSteps createAuthenticationSteps(
 		if ((roleId instanceof Wrapped || roleId instanceof Pull)
 				&& (secretId instanceof Wrapped || secretId instanceof Pull)) {
 
-			throw new IllegalArgumentException(
+			throw new VaultClientException(
```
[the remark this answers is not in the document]
mikebell90: This was one of the in between ones I guess could go either way. My only argument is hey if ones unifying up the hierarchy why ISN't this a VCE.

While I use IllegalState, IllegalArgument a lot, in an API, I'd argue unless it is a 100% obvious and expected consequence, it's best shielded, otherwise encouraging people to just catch RuntimeException.

But continuing on...

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -181,7 +182,7 @@ public static AuthenticationSteps createAuthenticationSteps(
 			});
 		}
 
-		throw new IllegalArgumentException(String.format(
+		throw new VaultClientException(String.format(
```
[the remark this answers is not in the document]
mikebell90: Same point back. But yeah. I'm not digging in btw, just explaining my rationale. If you feel strongly at the end, sure I'll change it.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -250,9 +251,9 @@ private String getRoleId(RoleId roleId) {
 				return (String) entity.getBody().getRequiredData().get("role_id");
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
```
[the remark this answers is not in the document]
mikebell90: Well, that was part of my original proposal.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -273,13 +274,13 @@ private String getRoleId(RoleId roleId) {
...
 						"Cannot unwrap Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 			}
 		}
 
-		throw new IllegalArgumentException("Unknown RoleId configuration: " + roleId);
+		throw new VaultClientException("Unknown RoleId configuration: " + roleId);
```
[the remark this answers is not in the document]
mikebell90: Reasonable. Upon looking this would mean things were whacked. Another option would be subclasses for each, and throw in the factory, but this probably fine. So agreed!

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -299,9 +300,9 @@ private String getSecretId(SecretId secretId) {
 				return (String) response.getRequiredData().get("secret_id");
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
```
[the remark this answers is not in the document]
mikebell90: I guess I'm not sure what you are arguing. There's an http code here. We could subclass VaultHttpException further to complain about VaultLogin, but only if we actually call a mapper that converts 403/401 appropriately. Is that what you are asking?

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -322,13 +323,13 @@ private String getSecretId(SecretId secretId) {
...
```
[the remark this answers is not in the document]
mikebell90: See above (back at ya ;) )

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
```diff
@@ -322,13 +323,13 @@ private String getSecretId(SecretId secretId) {
...
 						"Cannot unwrap Role id using AppRole: %s",
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
 			}
 		}
 
-		throw new IllegalArgumentException("Unknown SecretId configuration: " + secretId);
+		throw new VaultClientException("Unknown SecretId configuration: " + secretId);
```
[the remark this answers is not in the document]
mikebell90: Conceded that this is the same as 282.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
```diff
@@ -104,10 +105,13 @@ public VaultToken login() throws VaultException {
 				}
 			}
 			catch (HttpStatusCodeException e) {
-				throw new VaultException(String.format(
+				throw new VaultHttpException(String.format(
```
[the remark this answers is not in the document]
mikebell90: I'd love to, except, I'm not sure of the context now. What specifically are you arguing, that these could be mapped to even more specific exceptions?

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
```diff
@@ -104,10 +105,13 @@ public VaultToken login() throws VaultException {
...
 						"HTTP request %s in state %s failed with Status %s and body %s",
 						o, state, e.getStatusCode(),
-						VaultResponses.getError(e.getResponseBodyAsString())));
+						VaultResponses.getError(e.getResponseBodyAsString())), e.getStatusCode());
```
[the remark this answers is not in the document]
mikebell90: Guess I'm not getting any of your see aboves when it comes to converting to an HttpBaseException. Is your argument that they should be more specific? We don't know anything except vault returned an http code != 200. Oh and that there's a semi standardized interpretation in the C grade docs they provide ;)

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
```diff
@@ -104,10 +105,13 @@ public VaultToken login() throws VaultException {
...
+			}
+			catch (VaultException e) {
+				throw e;
 			}
 			catch (RuntimeException e) {
 				throw new VaultException(String.format(
```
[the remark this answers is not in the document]
mikebell90: True. I left it this way because it wasn't very clear at this level what had caused the issue, hence the generic exception level.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
```diff
@@ -126,7 +130,7 @@ public VaultToken login() throws VaultException {
 			return LoginTokenUtil.from(response.getAuth());
 		}
 
-		throw new IllegalStateException(String.format(
+		throw new VaultException(String.format(
```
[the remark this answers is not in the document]
mikebell90: But it is. You are running through a state machine and getting results unexpected, that normally would throw up. It's either a IllegalStateException or a VaultException, I grant you.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsOperator.java
```diff
@@ -127,13 +129,13 @@ public AuthenticationStepsOperator(AuthenticationSteps steps, WebClient webClien
 						return LoginTokenUtil.from(response.getAuth());
 					}
 
-					throw new IllegalStateException(
+					throw new VaultException(
```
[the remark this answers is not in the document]
mikebell90: You can't and have the errorMap make sense. It's gonna convert it anyway!

spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
mikebell90: Your point?

spring-vault-core/src/main/java/org/springframework/vault/authentication/IpAddressUserId.java
mikebell90: Yeah I was reserving VaultException as a catchall container root. This (arguably) is a VaultClientException because it occurs when on on the client side, something not to do with the specific call to Vault occurred.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
mikebell90: I'm really going to need clarification of these "see above". Remember this is the only one exception fits all and further subclasses can clarify. Because login isn't the only problem.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesServiceAccountTokenFile.java
mikebell90: It is, but ignores the reasons that VaultClientException was created, which we need to come to consensus on. It exists to clearly separate cause and effect
Cause: Something happened on the Client side while preparing a request = VaultClientException
Cause: Something happened while making the request = VaultRemoteException or VaultHttpException
Cause: Something happened while the request is done but parsing it initially: VaultRemoteException
Cause: It's definitely Vault Related but we've lost context - VaultException

That's my logic in a nutshell.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
mikebell90: Sure.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LoginTokenAdapter.java
mikebell90: I think what yiou are saying:
Embed the original HttpStatusCodeException in and remove the HttpCode assumption? Is this a general statement? Sorry I'm not following. I'm ok with embedding the original one, but how is the client to interpret it. I would say with a very rich strata of exceptions (VaultTokenLookup, VaultLoginFailed, etc). But originally in the Issues, you seemed to want to start small.

spring-vault-core/src/main/java/org/springframework/vault/authentication/MacAddressUserId.java
mikebell90: It's definitely an issue unquestionably on the client ;)

spring-vault-core/src/main/java/org/springframework/vault/authentication/MacAddressUserId.java
mikebell90: Getting a hardware address and it failed isn't client/local specific?

spring-vault-core/src/main/java/org/springframework/vault/authentication/VaultTokenLookupException.java
mikebell90: In what case is it not? Sorry just not following.

spring-vault-core/src/main/java/org/springframework/vault/client/VaultResponses.java
mikebell90: Sure.

spring-vault-core/src/main/java/org/springframework/vault/config/ClientHttpConnectorFactory.java
mikebell90: If anything is more local than specifying ones local ssl config, what is?

Fundamentally let's see if we agree on this:

There are exceptions that are local in origin, and from remote access.
That's the raison d'etre.

mikebell90:
So there's some fundamental disconnect that some back and forth should hopefully resolve. See my comments

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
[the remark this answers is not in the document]
mp911de: We're using IAE and ISE to distinguish between framework-specific issues and state/argument issues. There's not point in introducing own exceptions of a generic type that shadow the fact that IAE and ISE are usually programming bugs while `VaultException`s might be something to recover from.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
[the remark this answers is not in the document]
mp911de: In that particular area it makes sense to stick to your initial idea. As mentioned, we had to start somewhere and now we're specializing cases.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AppRoleAuthentication.java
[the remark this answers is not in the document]
mp911de: Sorry for not being clear. I meant to improve how the exception is created. Retaining the cause instead of propagating the status.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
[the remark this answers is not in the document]
mp911de: Having a `VaultLoginException` with the cause of `HttpStatusCodeException`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AuthenticationStepsExecutor.java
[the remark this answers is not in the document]
mp911de: Sorry, that's a duplicate.

spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsEc2Authentication.java
mp911de: This comment requires some context from `createTokenUsingAwsEc2()`: We're throwing `VaultHttpException` in `createTokenUsingAwsEc2()`. It bubbles up to the user without the context in which it happened (Login through AWS-EC2). It would make sense to catch runtime exceptions and wrap these into `VaultLoginException`.

Login failure handling become much more evident as it's a recurring pattern.

spring-vault-core/src/main/java/org/springframework/vault/authentication/IpAddressUserId.java
mp911de: Throwing/catching `VaultException` would typically result in just logging or application shutdown (in an extreme case). There's usually nothing you can do from an application perspective to recover from a `VaultException`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesAuthentication.java
mp911de: Was meant in the context of login to represent all top-level login failures within a `VaultLoginException`. Nested calls to external resources would retain `VaultHttpException`. Failures of calling `auth/{mount}/login` via HTTP or exceptions that bubble up otherwise would be wrapped within `VaultLoginException`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/KubernetesServiceAccountTokenFile.java
mp911de: We're catching `IOException` here for the sole purpose of wrapping it to a `RuntimeException`. There's nothing Vault specific here except that the code lives in Spring Vault.

mikebell90: Seems like that argument, if taken to extremes says "Convert everything to a generic RuntimeException, dump VaultException and VaultClientException and just wrap the HttpStatusException in a generic RuntimeException" ;)

I am pretty sure you don't mean that.
I'm making the case that this is middle ware and generally most exceptions should be "refunneled" into some subclass of VaultException.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LoginTokenAdapter.java
mp911de: True, the intention was to start small and not to change every single place we're throwing exceptions. I'm fine with the fact that it started growing and we can identify areas where more detailed exceptions make sense.

> Embed the original HttpStatusCodeException in and remove the HttpCode assumption? Is this a general statement?

Having a pull request is great as we see now how creating exception looks and feels and how the changes affect the code. We have exceptions that are purely HTTP related and exceptions that are caused by HTTP interaction but not necessarily bound to HTTP. That's why I mentioned as one of the last comments to use the cause for `VaultHttpException` to transport the actual cause so clients can inspect that if they are interested in that. We shouldn't duplicate functionality for the sake of improved exception cause iteration. 
One thing that isn't covered yet is timeouts and generic I/O errors which are subtypes of `RestClientException`, i.e. `ResourceAccessException`. They don't have a HTTP status yet are they HTTP-related.

If your goal is to inspect each and every HTTP status, then a `RestTemplate` interceptor seems to be the better fit.

> I would say with a very rich strata of exceptions

Very is relative to the observer. Introducing `VaultHttpException` and `VaultLoginException` seems reasonable small to me.

spring-vault-core/src/main/java/org/springframework/vault/authentication/MacAddressUserId.java
mp911de: The client is `RestTemplate` and `VaultTemplate`.

spring-vault-core/src/main/java/org/springframework/vault/authentication/VaultTokenLookupException.java
mp911de: In the case of I/O exceptions (such as read timeout, unresolvable host, too many redirects, SSL errors).

spring-vault-core/src/main/java/org/springframework/vault/config/ClientHttpConnectorFactory.java
mp911de: Agree there's a local/transport/remote distinction. If anything is not transport/remote-related, then it happens locally. No need to state the obvious to the user as long as we are consistent with the remote part.

mikebell90:
So here's what I recommend. Let's make this step by step.

1. My next commit basically puts the enclosed exception inside, which I probably should have done. I'll attach, you review.
2. Then let's add VaultLoginException separately.
3. Then let's come together and talk about our "disjoint" which is the VaultClientException, etc. Remaining. Let's take this last, so we can focus on progress.

Make sense?

mp911de:
This PR obviously became bigger than it should in the first place, mostly because it contains the new exceptions and changes the code where exceptions are thrown. I think this is fine because we can dig into detail what's actually happening where exceptions are thrown. From the initial ticket description it made sense to think in `VaultClientException` and `VaultHttpStatusException` categories. 

As we go forward, it does not make sense to put everything into only two categories, `VaultClientException` and say `VaultRemoteException`. We're using JDK exceptions and lines start to blur if we start thinking about e.g. login exception. Did they happen on the server? Did they occur on the client? Since both are possible, we should express the actual issue and not try to create a distinction between things that are hard to distinguish and do not enhance value on the client side.

What becomes obvious is that we need to deal with transport exception instead of HTTP status exceptions only. That's why I think we should have only a single exception for transport-related errors and allow introspection through the cause.

mp911de:
Agree on https://github.com/spring-projects/spring-vault/pull/216#issuecomment-372803416.

I'm also perfectly fine if we take point 3 out of scope and deal in a later pull request.

mikebell90:
Been horribly busy, will probably get back this weekend.

mp911de:
Any update on this one?

mikebell90:
I tried but failed to get back. Very work stressed sigh. To be honest I feel there is two choices: you do it or we restart with a much smaller scope initially. At this point I can’t even track who is what. I suppose that’s my fault for trying to do it all at once

mikebell90:
Hope that didn’t come out sulky. It is true I have been super busy but also true when I look at the plethora of comments I get frustrated at just where to start again

mp911de:
No worries. It wasn't ever my intent to upset you. I'd feel also more comfortable to close this one and start over. I can pick up loose ends on the first bits regarding login exceptions and then we can progress from there on.

WDYT?

mikebell90:
Makes sense to me

PR #219, opened 2018-03-12
Possible implementation of SessionManagerListener

mikebell90:
1. Right now I allow for renewal in the event. This allows for moving all of the code into there for renewal for the two LifecycleSessionManager, but is currently dubious logic.
2. My reactive fu is weak - so the Reactive part needs scrutiny.

Arguably instead of hitting the subclass we need separate implementations of listener in the two Lifecycle concrete classes to preserve reactive semantics. Then the renewal needed would return some form of TokenWrapper directly.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
@@ -19,8 +19,6 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
 
-import javax.swing.text.html.Option;
-
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
@@ -173,13 +173,6 @@ public class LifecycleAwareSessionManager extends LifecycleAwareSessionManagerSu
 	 */
 	protected boolean renewToken() {
 
-		Optional<VaultToken> listenerToken = listener.onSessionRenewalNeeded();
-		if (listenerToken.isPresent()) {
-			// Dubious logic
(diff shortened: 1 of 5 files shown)


mp911de:
Not quite sure about the idea behind the change. Could you please elaborate the rationale or what you're trying to achieve before proposing changes?

mikebell90:
Sure. Let me start by saying this is sort of a sketch and I'm not sold 100% on the solution.
Here are the problems I am trying to solve:

Primary: An Event that lets you know when the tokens are bad. 
Rationale: When wrapping spring vault into a service client (eg one of 1000 containers), it's important that a failure to renew tokens, etc is known and action can be done, such as logging metrics, exiting the container, firing alarms, debugging, etc. Right now that's all encapsulated and not expandable cleanly.

Secondary: Allow another option for getting a token. I guess I don't mind if we drop this per se, one could override renewToken. OTOH, there's an argument to be made that the entire renew/revoke logic itself should be attached to an event listener.

Improvement in signature: Instead of passing the token when renewal is successful, one could pass the SessionManager (which makes getting the token trivial). That would clean up the signature to 

onRenewedFailed(SessionManager)
onRenewedSuccessfully(SessionManager)

mikebell90:
So the primary purpose is to just allow an event driven set of hooks to react in any way seen fit to successful or failed renewals. It also parallels the SecretLeaseContainer

mp911de:
Thanks for the insight. We could provide hooks to allow event dispatching. `LifecycleAwareSessionManager` isn't as event-based as `SecretLeaseContainer`. We have to see whether a switch to an event-driven approach reduces the complexity or whether it raises complexity. 

We used events in `SecretLeaseContainer` to decouple things which helped to organize the code better.

mikebell90:
Right that's the basic idea. The secondary idea of switch it all to events is dubious, and if we dropped the first method (aboutToRenew) we could skip it for now and extend it later

mikebell90:
I've  now simplified this a lot
- Removed the problematic method allowing overriding of Token acquisition
- Both remaining methods provide no injection of parameters.

mikebell90:
@mp911de I think this is now simple enough to be clear and extensible ;)

mp911de:
We decided to not merge the pull request as we feel there's not sufficient demand for the feature. Thanks a lot for contributing.

Issue #223, opened 2018-03-20
Batch decryption fails if plaintext is empty


mikkokoli:
I have a bunch of ciphertexts I need to decrypt in a batch for performance reasons. Some of the ciphertexts decrypt into an empty string. 

Trying to do this with `TransitOperations` fails. It tries to create a Plaintext, but `Plaintext.of()` expects a non empty string.

In my opinion `Plaintext.of()` should accept an empty string. Vault itself will happily accept empty strings for the plaintext.

mp911de:
We should align to what Vault does. It may make sense for security reasons to encrypt empty values (see #1874).

mp911de:
I reopened this one to fix the subsequent issue. Feel free to submit a pull request.

PR #225, opened 2018-03-21
Fixes gh-223 allow empty plaintext for transit decryption

mikkokoli:


mp911de:
Thanks a lot. That's merged, polished and backported now.

Issue #243, opened 2018-04-18
VaultTokenRequest noParent field serializes incorrectly in JSON


voberle:
In https://github.com/spring-projects/spring-vault/blob/master/spring-vault-core/src/main/java/org/springframework/vault/support/VaultTokenRequest.java
the `noParent` field is missing `@JsonProperty("no_parent")`

Without it, it becomes `noParent` in the JSON, but Vault expects `no_parent`
https://www.vaultproject.io/api/auth/token/index.html#no_parent

voberle:
```
--- a/spring-vault-core/src/main/java/org/springframework/vault/support/VaultTokenRequest.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/support/VaultTokenRequest.java
@@ -42,6 +42,7 @@ public class VaultTokenRequest {
 
        private final Map<String, String> meta;
 
+       @JsonProperty("no_parent")
        private final boolean noParent;
 
        @JsonProperty("no_default_policy")
```

mp911de:
Good catch! Care to submit a pull request?

voberle:
Here is the PR https://github.com/spring-projects/spring-vault/pull/244

PR #244, opened 2018-04-18
Fix JSON serialization of VaultTokenRequest noParent (#243)

voberle:
In `VaultTokenRequest.java` the `noParent` field is missing `@JsonProperty("no_parent")`

Without it, it becomes `noParent` in the JSON, but Vault expects `no_parent`
https://www.vaultproject.io/api/auth/token/index.html#no_parent


mp911de:
Thanks a lot. That's now merged and backported.

Issue #262, opened 2018-07-18
Update sys/ calls to use newer APIs for vault 0.8+


baens:
It looks like [Vault 0.8](https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#080-august-9th-2017) deprecated some of their sys API end points, especially ones that have to renew them.

Does there need to be a way to configure where the endpoints are, or just update them in place?

mp911de:
The mentioned endpoints are used only in `SecretLeaseContainer` and are not configurable yet. We could introduce a lease accessor/lease endpoint strategy that allows mapping the changes which are bound to versions pre 0.7.3/post 0.8. Right now, we only use `revoke` and `renew` operations.

baens:
Correct the `SecretLeaseContainer` was the specific class I saw that needed to be changed.

Just to make sure I understand, the `SecretLeaseContainer` is the class that handles the renewal of database credentials or other tokens that needed to be renewed periodically correct? Or is there somewhere else that also handles this? (i.e. where does the `LifecycleAwareSessionManager` come into play)

I would love to contribute, are there any examples in this code base of how to apply the strategy pattern in the way you want here?

mp911de:
> Just to make sure I understand, the SecretLeaseContainer is the class that handles the renewal of database credentials.

That's correct.

> where does the `LifecycleAwareSessionManager` come into play and how does it differ from `SecretLeaseContainer`

`LifecycleAwareSessionManager` uses `auth/token/revoke-self` and `auth/token/renew-self` API calls to renew session tokens. These endpoints aren't deprecated so it's safe to leave `LifecycleAwareSessionManager` as-is. `SecretLeaseContainer` obtains secrets from secret backends which can be associated with a lease, hence it's the one place that deals with renewal and revocation of secret leases.

> I would love to contribute

Great to hear.

> are there any examples in this code base of how to apply the strategy pattern in the way you want here?

I think no, there isn't yet code that compares well to what we plan here.

I'd imagine having an interface (`LeaseAccessor`), that declares two methods:
* `revoke(Lease, RestOperationsCallback<Void>)`
* `renew(Lease, RestOperationsCallback<Lease>)`

And then we could declare an enum that implements the interface. The enum would have two items (`SysLease` and `Legacy`, not exactly sure about the name) and each implements its own API access scheme along with paths and request/response handling. This way, we encapsulate the entire endpoint calls within the corresponding method and the endpoint calls do not share details about request/responses.

[...]

PR #282, opened 2018-08-02
Moving lease data into object in lease PUT

baens:
It appears that maybe this URL just needs to be passed an object instead of having the `lease_id` embedded in the URL and that solves most of the problems I have.

The URL is still deprecated, but this solves the immediate problem that the URL with the embedded lease_id no longer has default permissions.

This will fix Issue #262.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
@@ -35,6 +35,7 @@ import lombok.extern.apachecommons.CommonsLog;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@baens Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=282)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-vault&pullRequestId=282) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@baens Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=282)!

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
```diff
@@ -592,11 +588,16 @@ protected Lease doRenewLease(RequestedSecret requestedSecret, Lease lease) {
 
 	@SuppressWarnings("unchecked")
 	private Lease renew(Lease lease) {
+		final Map<String, String> leaseRenewalData = new LinkedHashMap<>();
```
baens: I was looking for an ImmutableMap of some kind, but the Guava version was failing. I would be more than happy to change this

spring-vault-core/src/main/java/org/springframework/vault/core/lease/SecretLeaseContainer.java
```diff
@@ -592,11 +593,16 @@ protected Lease doRenewLease(RequestedSecret requestedSecret, Lease lease) {
 
 	@SuppressWarnings("unchecked")
 	private Lease renew(Lease lease) {
+		final Map<String, String> leaseRenewalData = new HashMap<>();
```
mp911de: Typically, we avoid `final` in method bodies since our upgrade to Java 1.8. I'll take care of this during the merge.

RE your comment on immutable maps: We also do not require immutable maps as the map does not leave this method.

baens: Anything else you want me to take care of or is this ok?

baens:
@mp911de Anything else you would like to see from this? Or will this work?

mp911de:
We'll take care of this pull request. Currently busy with other things.

mp911de:
That's merged and polished now. Thanks for your contribution.

Issue #336, opened 2018-11-22
Upgrade to Vault 1.0.0-rc1


DanielFran:
See https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#100-rc1-nov-20th-2018

PR #337, opened 2018-11-22
Upgrade to Vault 1.0.0-rc1

DanielFran:
See https://github.com/spring-projects/spring-vault/issues/336

### DIFF
--- a/src/test/bash/install_vault.sh
+++ b/src/test/bash/install_vault.sh
@@ -6,7 +6,7 @@
 ###########################################################################
 
 
-VAULT_VER="${VAULT_VER:-1.0.0-beta2}"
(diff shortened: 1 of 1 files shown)


DanielFran:
The failing case is due to this change:
_Paths within auth/token that allow specifying a token or accessor in the URL have been removed. These have been deprecated since March 2016 and undocumented, but were retained for backwards compatibility. They shouldn't be used due to the possibility of those paths being logged, so at this point they are simply being removed._

What is the expected in that situation?

mp911de:
Thanks a lot. We should pass the token in the `POST` body. I'll take care of this during the merge.

mp911de:
That's merged, polished, and backported now.

PR #434, opened 2019-06-07
Upgrade to Vault 1.1.3

DanielFran:
See https://github.com/hashicorp/vault/blob/v1.1.3/CHANGELOG.md


mp911de:

Issue #433, opened 2019-06-07
VaultMount does not allow for options to be specified


maciejqantas:
Affected version: 2.1.2

To properly create a KV v2 versioned secret backend, options must be specified as per https://www.vaultproject.io/api/system/mounts.html#options.
Also, the link above incorrectly shows that `options` should be a field of `config`. It is a separate field, as per https://github.com/hashicorp/vault/blob/master/api/sys_mounts.go#L137. A fix has been requested in https://github.com/hashicorp/vault/issues/4601#issuecomment-500802035.

Currently, `VaultMount` does not allow for a specification of `options`, only `config` is exposed. This way, if a mount of `type` set to `kv` is created, it's not versioned and versioning must be enabled separately.

A workaround is to use `VaultTemplate` directly (as opposed to `VaultSysOperations`) and execute `write("sys/mounts/mymount", "{\"type\": \"kv\", \"options\": {\"version\": \"2\"}}")`.

mp911de:
That's a good enhancement proposal. Do you want to submit a pull request?

maciejqantas:
Sure, I'll give it a go soon.

maciejqantas:
@mp911de I will need your help here.
Tests are failing for older Vault versions. Is there a mechanism to run tests for specific vault versions only?

mp911de:
You already answered your first question. No need for a separate test class. If there's no version check yet, then add it like it is done in https://github.com/spring-projects/spring-vault/blob/c7b6557606b7e3a41a7982326612615c192adff1/spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java#L55-L62.

maciejqantas:
@mp911de thanks for that! I didn't even realise it was in the same class I worked on 😊 

Travis seems happy, have a read of the code please.

PR #435, opened 2019-06-12
Add options to VaultMount

maciejqantas:
Fixes https://github.com/spring-projects/spring-vault/issues/433.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/support/VaultMount.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/support/VaultMount.java
@@ -50,13 +50,20 @@ public class VaultMount {
 	 */
 	private final Map<String, Object> config;
 
+	/**
+	 * Mount type specific options.
+	 */
+	private final Map<String, String> options;
+
 	VaultMount(@JsonProperty("type") String type,
 			@Nullable @JsonProperty("description") String description,
-			@Nullable @JsonProperty("config") Map<String, Object> config) {
+			@Nullable @JsonProperty("config") Map<String, Object> config,
+			@Nullable @JsonProperty("options") Map<String, String> options) {
 
 		this.type = type;
 		this.description = description;
 		this.config = config != null ? config : Collections.emptyMap();
+		this.options = options != null ? options : Collections.emptyMap();
 	}
 
 	/**
@@ -100,6 +107,14 @@ public class VaultMount {
 	}
 
 	/**
+	 * @return mount type specific options.
+	 */
+	@Nullable
+	public Map<String, String> getOptions() {
+		return options;
+	}
+
+	/**
 	 * Builder to build a {@link VaultMount}.
 	 */
 	public static class VaultMountBuilder {
@@ -111,6 +126,8 @@ public class VaultMount {
 		private String description;
 
 		private Map<String, Object> config = Collections.emptyMap();
+
+		private Map<String, String> options = Collections.emptyMap();
 
 		VaultMountBuilder() {
 		}
@@ -156,6 +173,20 @@ public class VaultMount {
 		}
 
 		/**
+		 * Set mount type specific options for this mount.
+		 *
+		 * @param options mount type specific options for this mount.
+		 * @return {@literal this} {@link VaultMountBuilder}.
+		 */
+		public VaultMountBuilder options(Map<String, String> options) {
+
+			Assert.notNull(options, "Options map must not be null");
+
+			this.options = options;
+			return this;
+		}
+
+		/**
 		 * Build a new {@link VaultMount} instance. Requires {@link #type(String)} to be
 		 * configured.
 		 *
@@ -166,7 +197,7 @@ public class VaultMount {
 			Assert.notNull(type, "Type must not be null");
 			Assert.hasText(type, "Type must not be empty or null");
 
-			return new VaultMount(type, description, config);
+			return new VaultMount(type, description, config, options);
 		}
 	}
 }
--- a/spring-vault-core/src/main/java/org/springframework/vault/support/VaultMount.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/support/VaultMount.java
@@ -108,6 +108,8 @@ public class VaultMount {
 
 	/**
 	 * @return mount type specific options.
+	 *
+	 * @since 2.2
 	 */
 	@Nullable
 	public Map<String, String> getOptions() {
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@maciejqantas Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=435)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-vault&pullRequestId=435) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@maciejqantas Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=435)!

mp911de:
See also #433.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java
```diff
@@ -98,6 +99,78 @@ public void mountShouldMountGenericSecret() {
...
+
+		try {
+			// a versioned write (kv put) will fail for a kv (default version: 1, not versioned) store
+			vaultOperations.opsForVersionedKeyValue("other").put("secret",
+				Collections.singletonMap("key", "value"));
+
+			fail("Missing NullPointerException");
+		}
+		catch (NullPointerException e) {
+			assertThat(e).hasMessage(null);
```
mp911de: This is odd. Seeing a catch for `NullPointerException` rather requires a fix than an assertion.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java
```diff
@@ -98,6 +99,78 @@ public void mountShouldMountGenericSecret() {
 		assertThat(Arrays.asList("kv", "generic")).contains(secret.getType());
 	}
 
+	@Test
+	public void mountShouldMountKv1Secret() {
+
+		assumeTrue(vaultVersion.isGreaterThanOrEqualTo(
+			VaultRule.VERSIONING_INTRODUCED_WITH));
+
+		if (adminOperations.getMounts().containsKey("other/")) {
```
mp911de: Let's use a more meaningful name as state persists beyond test runs and we should get an indication where this mount comes from.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultMount.java
```diff
@@ -155,6 +172,20 @@ public VaultMountBuilder config(Map<String, Object> config) {
 			return this;
 		}
 
+		/**
+		 * Set mount type specific options for this mount.
+		 *
+		 * @param options mount type specific options for this mount.
+		 * @return {@literal this} {@link VaultMountBuilder}.
```
mp911de: Nit: Add `@since` tag.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultMount.java
```diff
@@ -99,6 +106,14 @@ public String getDescription() {
 		return config;
 	}
 
+	/**
+	 * @return mount type specific options.
+	 */
+	@Nullable
+	public Map<String, String> getOptions() {
```
mp911de: Nit: Add `@since` tag.

pom.xml
```diff
@@ -415,7 +415,7 @@
 			<id>springNext</id>
 
 			<properties>
-				<spring.version>5.1.5.BUILD-SNAPSHOT</spring.version>
+				<spring.version>5.2.0.BUILD-SNAPSHOT</spring.version>
```
mp911de: Let's keep dependency changes out of this PR.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java
maciejqantas: It is weird indeed, but I wasn't sure how to approach it.

The `put` call ends up returning no response body - it's a 204, and 3 headers are returned. As a result, an NPE comes from the `response` being `null` in `return getMetadata(response.getRequiredData());` in https://github.com/spring-projects/spring-vault/blob/master/spring-vault-core/src/main/java/org/springframework/vault/core/VaultVersionedKeyValueTemplate.java#L155.

There's nothing to build `Metadata` from. Would you like me to return an empty `Metadata` object in that case?

spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java
maciejqantas: Sure, will do.

pom.xml
maciejqantas: I did it because of https://travis-ci.org/spring-projects/spring-vault/jobs/544499235.

This line is for the `springNext` profile only. Moreover, in line 29, we already have `<spring.version>5.2.0.M2</spring.version>`. I figured `springNext` should at least be a higher version, or the same one if there are no higher versions.

Happy to revert if that's what you want though.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java
mp911de: So there are two things here: Using `opsForVersionedKeyValue` with `kv` version 1 is wrong per se because it's the wrong backend. In any case we should introduce a check and throw `IllegalStateException` if the response is null.  

For this test I suggest using the regular `VaultTemplate.write`/`read` methods. Seeing `opsForVersionedKeyValue`
with kv version 1 is misleading.

pom.xml
mp911de: I see. I fixed that issue via 7460d5058b08baa2cf237706d2635f8a9fc27d8c.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultSysTemplateIntegrationTests.java
maciejqantas: @mp911de understood. My angle was such that one would expect kv-specific operations to work for a `kv` mount.
Will implemented as requested.

maciejqantas:
@mp911de thanks for your feedback. I've addressed all your comments.

mp911de:

PR #450, opened 2019-07-27
Update Vault version to 1.1.4

DanielFran:
See https://github.com/hashicorp/vault/blob/v1.1.4/CHANGELOG.md


mp911de:

PR #452, opened 2019-07-31
Update Vault version to 1.1.5 and 1.2.0

DanielFran:
See https://github.com/hashicorp/vault/blob/v1.1.5/CHANGELOG.md


mp911de:
How about switching to Vault 1.2.0?

DanielFran:
OK, will push changes

mp911de:

PR #476, opened 2019-09-18
Added ability to configure uri_sans when building issue certificate request

aantono:
Vault API provides a parameter to configure `uri_sans` when calling to request a certificate issue - https://www.vaultproject.io/api/secret/pki/index.html#uri_sans-1

Added ability to configure such parameter in the request builder.


mp911de:
Thanks a lot, that's merged and polished now.

PR #485, opened 2019-09-30
Upgrade Vault to 1.2.3.

DanielFran:
See https://github.com/hashicorp/vault/blob/v1.2.3/CHANGELOG.md


mp911de:

PR #505, opened 2019-11-09
Use curl instead of wget in install_vault.sh

bdemers:
curl is installed by default on many systems


mp911de:

Issue #462, opened 2019-08-22
Add authentication path customization in EnvironmentVaultConfiguration


hitenny:
I am using spring vault to access Vault from Spring boot app running in Kubernetes. 

**Version**

    <dependency>  
           <groupId>org.springframework.vault</groupId>  
           <artifactId>spring-vault-core</artifactId>  
           <version>2.1.3.RELEASE</version>  
    </dependency>
  

**Config**  

    vault:
      uri: https://xxx.xxx.com:8200
      authentication: KUBERNETES
      kubernetes:
        role: abc
        kubernetes_path: path/to/k8s
        service_account_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token

**Error** 

    o.s.v.a.VaultLoginException: Cannot login using Kubernetes: invalid role name \"abc\";

When I try to login using curl with the same role and token, its success:

    VAULT_LOGIN="{\"role\":\"$SA_ROLE\", \"jwt\":\"$SA_JWT_TOKEN\"}"
    curl --request POST --data "$VAULT_LOGIN" https://xxx.xxx.com:8200/v1/auth/path/to/k8s/login

mp911de:
Can you make sure to use Apache HttpClient and increase the logging level to enable wire logging to inspect the outgoing payload for the login?

mp911de:
Looking at the implementation code, the slashes in the given path are URL-escaped (`%2F` instead of `/`). All authentication implementations suffer from that issue.

hitenny:
I have analyzed the spring-vault code and found an issue here: https://github.com/spring-projects/spring-vault/blob/ba7fd9b8b461f1bb52cbd74ae234079eac2fccef/spring-vault-core/src/main/java/org/springframework/vault/config/EnvironmentVaultConfiguration.java#L357

it does not read path from the configuration. I have fixed in my application by overriding this method: `protected ClientAuthentication kubeAuthentication()` in my application

hitenny:
> Can you make sure to use Apache HttpClient and increase the logging level to enable wire logging to inspect the outgoing payload for the login?

I logged the REST api call which pring-vault makes to Vault for login. Irrespective of the path configured in yaml, it is sending the default path `KUBERNETES` to vault. Root cause is provided in my previous comment

mp911de:
Thanks a lot. Can you file a bug report in Spring Cloud Vault? We cannot move tickets across organizations. 

Closing this one as the issue is in Spring Cloud Vault.

[...]

PR #507, opened 2019-11-11
provide configurable custom mount path for vault auth backends

raoofm:


raoofm:
@mp911de please review. Can this be back-ported to 2.1 as well

raoofm:
fixes #462

mp911de:
We should be okay with backporting that change to the 2.2 line but not 2.1. We usually do not backport features to avoid all sorts of downstream issues, especially as this functionality can easily go unnoticed as we optionally pick up values from the environment.

PR #511, opened 2019-11-26
if response status from vault is 404, then return Mono.empty for read/list and Mono.error for write/delete

### LINKED ISSUE #503: write call is returning Mono.empty() when request failed with 404 Not Found
Our app tried to write to vault and vault returned 404 because the path doesn't exist.

**Expected behaviour**: We assumed in such scenario spring-vault would result mono error

**Actual behaviour**: it actually returns Mono.empty().

Have a look at this method in vault/core/ReactiveVaultTemplate.java

```
	private static <T> Mono<? extends T> mapOtherwise(ClientResponse response, String path) {

		if (response.statusCode() == HttpStatus.NOT_FOUND) {
			return Mono.empty();
		}

```

NOTE: the problem is when vault returns 204 No Content, spring-vault also results in Mono.empty(), which makes it almost impossible for downstream to handle 204 and 404 responses.

raoofm:
fixes #503

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/ReactiveVaultTemplate.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/ReactiveVaultTemplate.java
@@ -24,6 +24,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import org.springframework.core.ParameterizedTypeReference;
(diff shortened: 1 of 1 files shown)


raoofm:
@mp911de please review

raoofm:
ok the tests passed now. @mp911de ready for review

mp911de:

Issue #532, opened 2020-01-29
Authentication {mount} paths are URL-encoded


alehlapatsin:
The following template is being used in KubernetesAuthentication.java: `auth/{mount}/login`
The problem if `kubernetes-path` is set to `hello/world`, the result path will be `auth/hello%2Fworld/login`, which is not correct.

mp911de:
Thanks for raising this issue. We face the same issue in other authentication methods. Care to address these in your pull request #533 as well? The more relevant changes we can collect into a release the earlier we can ship a bugfix release.

alehlapatsin:
@mp911de I updated the PR with the other fixes. Could you please take a look?

PR #533, opened 2020-01-29
Mount path shouldn't be escaped

alehlapatsin:
Resolves KubernetesAuthentication issue, where the mount point is being escaped. 

The link to the issue: https://github.com/spring-projects/spring-vault/issues/532


alehlapatsin:
@mp911de Thank you for looking into the PR. I updated the PR with the other fixes.

mp911de:

PR #542, opened 2020-03-09
Expanded EnvironmentVaultConfiguration options for AzureMSIAuthentication

justbert:
I've run into an issue where my Identity Token Service URI isn't the default. It's not a huge deal, but it means that I'm unable to use the EnvironmentVaultConfiguration to simply setup my Boot project.

I believe that this slight tweak which uses all of the default functionality already present, but allows for a simpler and faster configuration through properties, may be of use to others. I'm definitely looking forward to being able to use it at work.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/config/EnvironmentVaultConfiguration.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/config/EnvironmentVaultConfiguration.java
@@ -145,6 +145,10 @@ import org.springframework.web.client.RestOperations;
  * <li>Azure MSI path: {@code vault.azure-msi.azure-path} (since 2.2.1, defaults to
  * {@link AzureMsiAuthenticationOptions#DEFAULT_AZURE_AUTHENTICATION_PATH})</li>
  * <li>Role: {@code vault.azure-msi.role}</li>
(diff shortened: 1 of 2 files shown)


mp911de:
Thanks for your pull request. It makes sense to extend the Environment configuration. Can you help us understand in which cases the URIs are different? While building the initial support, it wasn't clear whether changing the URLs was a requirement.

mp911de:

justbert:
Perfect thank you very much!

I'm still trying to figure it out all of the cases (the docs are a bit obtuse).
In my case, the Identity Token Service doesn't have a resource registered (or it isn't responding) for https://vault.hashicorp.com. I ended up using https://azure.manage.com to retrieve the identity token which works and which I have seen for most examples I can find online.

I've also updated both the identity service and the metadata service URIs to use the latest api-versions (2019-11-01) which may not be necessary, but I'm not positive on Azure's deprecation policies.

justbert:
PS, I appreciate the cleanup, wasn't very sure on all the conventions or how far reaching my changes could range!

mp911de:
No worries. We're more than happy to pick up on first-timer contributions if they make sense and apply a round of polishing. This way, you can learn about our style for future contributions.

Issue #432, opened 2019-06-05
No ops for deleting metadata in Vault template


zak905:
It would be nice to have operations for deleting and getting metadata implemented for `kv` (and same thing for secrets I suppose), like the ones specified by vault CLI:
```
vault kv metadata
Usage: vault kv metadata <subcommand> [options] [args]

  This command has subcommands for interacting with the metadata endpoint in
  Vault's key-value store. Here are some simple examples, and more detailed
  examples are available in the subcommands or the documentation.

  Create or update a metadata entry for a key: 

      $ vault kv metadata put -max-versions=5 secret/foo

  Get the metadata for a key, this provides information about each existing
  version:

      $ vault kv metadata get secret/foo

  Delete a key and all existing versions:

      $ vault kv metadata delete secret/foo

  Please see the individual subcommand help for detailed usage information.

Subcommands:
    delete    Deletes all versions and metadata for a key in the KV store
    get       Retrieves key metadata from the KV store
    put       Sets or updates key settings in the KV store
```

The reason why metadata operations are important is because attempting to delete a key will only delete the data it contains, and the key will still be there. So to delete a key completely, the metadata needs to be deleted. As a workaround right now, I will be using rest API call directly, anyways, if this is of interest, I could submit a PR eventually, it could take some time however.

mp911de:
Thanks for the ticket. That's a good enhancement. Pull requests are welcome. We should probably introduce a dedicated interface (`VaultKeyValueMetadataOperations`) that can be obtained from `VaultVersionedKeyValueOperations` to keep `VaultVersionedKeyValueOperations` focussed on key operations.

PR #561, opened 2020-05-19
add operations for interacting with kv metadata

zak905:
as promised in #432, here is the functionality for interacting with a kv metadata. Looking forward to hear your feedback.


pivotal-issuemaster:
@zak905 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=561)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-vault&pullRequestId=561) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@zak905 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=561)!

spring-vault-core/src/test/java/org/springframework/vault/util/VaultInitializer.java
```diff
@@ -96,7 +96,7 @@ public void initialize() {
 
 			if (this.prepareVault.getVersion()
 					.isGreaterThanOrEqualTo(VERSIONING_INTRODUCED_WITH)) {
-				this.prepareVault.disableGenericVersioning();
+				//this.prepareVault.disableGenericVersioning();
```
zak905, line 99: I know this is not the right fix, but I was just wondering why using `/versioned` path and unmounting the already provided `/secret`. In real world settings, `/secret` is the default right?

mp911de: This is true if you look at it from a most recent version perspective. We're running our test against various Vault versions that do not contain the KV version 2 backend so we're creating a stable setup. Also, newer Vault versions were used to mount the versioned backend at `/secret` so we need to revert this change.

spring-vault-core/pom.xml
```diff
@@ -75,6 +75,11 @@
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>com.fasterxml.jackson.datatype</groupId>
```
mp911de, line 79: We should rather not pull in the JSR-310 support here as this dependency change has a few downstream consequences (such as customizing the `ObjectMapper`).

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValueMetadataTemplate.java
```diff
@@ -0,0 +1,82 @@
...
+  static {
+    ObjectMapper mapper = new ObjectMapper();
+    JavaTimeModule module = new JavaTimeModule();
+    mapper.registerModule(module);
+    mapper.configure(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE, true);
+    OBJECT_MAPPER = mapper;
+  }
+
+
+  public VaultKeyValueMetadataTemplate(VaultOperations vaultOperations) {
```
mp911de, line 32: We need to pass on the path for the kv backend, similar to `VaultVersionedKeyValueTemplate`.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValueMetadataTemplate.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+public class VaultKeyValueMetadataTemplate implements VaultKeyValueMetadataOperations {
+
+  private final VaultOperations vaultOperations;
+
+  private static final ObjectMapper OBJECT_MAPPER;
+
+  static {
+    ObjectMapper mapper = new ObjectMapper();
```
mp911de, line 24: Given the need for another dependency, we should rather parse JSR-310 dates ourselves (see `VaultVersionedKeyValueTemplate.getDate(…)`)

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValueMetadataTemplate.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public VaultKeyValueMetadataTemplate(VaultOperations vaultOperations) {
+    Assert.notNull(vaultOperations, "VaultOperations must not be null");
+    this.vaultOperations = vaultOperations;
+  }
+
+  @Override
+  public void delete(String path) {
+    Assert.hasText(path, "Path must not be empty");
+    vaultOperations.doWithSession(restOperations -> {
```
mp911de, line 40: `vaultOperations.delete(…)` should be sufficient.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValueMetadataTemplate.java
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }
+
+  @Override
+  public void put(String path, VaultMetadataRequest body) {
+    Assert.hasText(path, "Path must not be empty");
+    Assert.notNull(body, "Body must not be null");
+    vaultOperations.doWithSession(restOperations -> {
+      try {
+        restOperations.put("/secret/metadata/" + path, body);
```
mp911de, line 74: `vaultOperations.write(…)` should be sufficient.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultOperations.java
```diff
@@ -54,6 +54,13 @@
 	 */
 	VaultKeyValueOperations opsForKeyValue(String path, KeyValueBackend apiVersion);
 
+	/**
+	 * Return {@link VaultKeyValueMetadataOperations}
+	 *
+	 * @return the operations interface to interact with the Vault Key/Value metadata backend
+	 */
+	VaultKeyValueMetadataOperations opsForKeyValueMetadata();
```
mp911de, line 62: Metadata operations should be obtained from `VaultVersionedKeyValueOperations.opsForMetatada()` or `.metadata()`. This way, we preserve the secret backend path and put the operations interface into an area where it belongs to.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultMetadataResponse.java
```diff
@@ -0,0 +1,184 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * Value object to bind Vault HTTP kv read metadata API responses.
+ *
+ * @author Zakaria Amine
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class VaultMetadataResponse {
```
mp911de, line 15: As mentioned earlier, let's perform deserialization of this type ourselves. That way, we get the most control over the object and we can reuse `Versioned.Metadata` for `versions` instead of exposing a Jackson-specific type.

We should make this type also immutable because the response object does not change once constructed.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultMetadataResponseWrapper.java
```diff
@@ -0,0 +1,17 @@
+package org.springframework.vault.support;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class VaultMetadataResponseWrapper {
```
mp911de, line 6: With custom deserialization, there's no need for this type. An inner class (`VaultMetadataResponseWrapper 
 extends VaultResponseSupport<Map<String, Object>>`) in `VaultKeyValueMetadataTemplate` should be sufficient.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueMetadataTemplateIntegrationTests.java
```diff
@@ -0,0 +1,111 @@
...
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = VaultIntegrationTestConfiguration.class)
+public class VaultKeyValueMetadataTemplateIntegrationTests extends AbstractVaultKeyValueTemplateIntegrationTests {
+
+  private static final String SECRET_NAME = "test";
+  private VaultKeyValueMetadataOperations vaultKeyValueMetadataOperations;
+
+  VaultKeyValueMetadataTemplateIntegrationTests() {
+    super("secret", VaultKeyValueOperationsSupport.KeyValueBackend.versioned());
```
mp911de, line 34: This path should be `versioned` for our arrangement.

spring-vault-dependencies/pom.xml
```diff
@@ -86,6 +86,12 @@
 				<version>${jackson-databind.version}</version>
 			</dependency>
 
+			<dependency>
```
mp911de, line 89: Should be removed as per comment in the other pom.xml.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueMetadataTemplateIntegrationTests.java
```diff
@@ -0,0 +1,111 @@
...
+import org.springframework.vault.support.VaultMetadataResponse;
+import org.springframework.vault.support.VaultResponse;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = VaultIntegrationTestConfiguration.class)
+public class VaultKeyValueMetadataTemplateIntegrationTests extends AbstractVaultKeyValueTemplateIntegrationTests {
```
mp911de, line 28: Nit: This test should be guarded so it does not run on Vault versions prior to 0.10.0 (`@RequiresVaultVersion(VaultInitializer.VERSIONING_INTRODUCED_WITH_VALUE)`)

zak905:
Hey, thanks for the code review. I did a quick skim through the comments. I am gonna address them soon.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValueMetadataTemplate.java
zak905: it seems like `write` does a POST while here we need a PUT

spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueMetadataTemplateIntegrationTests.java
zak905: The annotation is already on the parent class, should I put it here as well?

zak905:
@mp911de I addressed pretty much all the comments, happy to hear your feedback.

mp911de:
Thanks a lot. I'm going to take the pull request from here.

mp911de:

zak905:
Welcome! Thanks.

Issue #585, opened 2020-09-23
Support for KV Patch in VaultTemplate


yfaney:
Hi, our team needs a KV Patch operation in Spring Boot and I was wondering if we can get a feature of KV Patch operation in VaultTemplate or not.

Here is the CLI command we use:

```bash
vault kv patch secret/ourteam/config foo=bar
```

mp911de:
Good proposal. Do you want to submit a pull request?

yfaney:
Sure, I'll kick the work off and post its pr.

PR #587, opened 2020-09-24
Adding support of KV Patch operation

yfaney:
# Issue Link
[Support for KV Patch in VaultTemplate #585](https://github.com/spring-projects/spring-vault/issues/585)

# Motivation and Context 
This change will enable the support of KV patch operation in VaultKeyValue2Template.
Note that this command will only work with Vault Secret Engine V2.
During the change, I tweaked a doRead() method to get the metadata properly in V2 access which is required to perform patch operation.
Please feel free to suggest improvement of this code. I tried the minimum change so it may have a room to refactor.
This change is based on the source code of the vault CLI:
https://github.com/hashicorp/vault/blob/master/command/kv_patch.go#L66

# How to test this change
Performing mvn install

# Test Evidence
Please see the attached screenshot
Screen Shot 2020-09-24 at 12 52 04 AM

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
@@ -16,10 +16,12 @@
 package org.springframework.vault.core;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.vault.VaultException;
 import org.springframework.vault.support.VaultResponse;
 import org.springframework.vault.support.VaultResponseSupport;
 
@@ -87,6 +89,40 @@ class VaultKeyValue2Template extends VaultKeyValue2Accessor implements VaultKeyV
 		Assert.hasText(path, "Path must not be empty");
 
 		doWrite(createDataPath(path), Collections.singletonMap("data", body));
+	}
+
+	/**
+	 * Performs a KV Patch operation.
+	 * @param path must not be {@literal null} or empty.
+	 * @param kv New key value map to be updated
+	 * @since 2.3
+	 */
+	@Override
+	public boolean patch(String path, Map<String, ?> kv) {
+		Assert.hasText(path, "Path must not be empty");
+
+		// To do patch operation, we need to do a read operation first
+		VaultResponse readResponse = get(path);
+		if (null == readResponse) {
+			throw new VaultException("VaultResponse must not be null");
+		} else if (null == readResponse.getData()) {
+			throw new SecretNotFoundException("No data found at %s; patch only works on existing data");
+		} else if (null == readResponse.getMetadata()) {
+			throw new VaultException("Metadata must not be null");
+		}
+		Map<String, Object> data = readResponse.getData();
+		Map<String, Object> metadata = readResponse.getMetadata();
+		kv.forEach(data::put);
+		Map<String, Object> body = new HashMap<>();
+		body.put("data", data);
+		body.put("options", Collections.singletonMap("cas", metadata.get("version")));
+
+		VaultResponse writeResponse = doWrite(createDataPath(path), body);
+		if (null == writeResponse) {
+			return false;
+		}
+		Map<String, Object> writeResponseData = writeResponse.getData();
+		return null != writeResponseData;
 	}
 
 }
--- a/spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueTemplateVersionedIntegrationTests.java
+++ b/spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueTemplateVersionedIntegrationTests.java
@@ -15,11 +15,22 @@
  */
 package org.springframework.vault.core;
 
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.vault.core.VaultKeyValueOperationsSupport.KeyValueBackend;
+import org.springframework.vault.support.VaultResponse;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * Integration tests for {@link VaultKeyValue2Template} using the versioned Key/Value (k/v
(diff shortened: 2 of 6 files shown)


pivotal-issuemaster:
@yfaney Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=587)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-vault&pullRequestId=587) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yfaney Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-vault&pullRequestId=587)!

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
```diff
@@ -89,4 +90,29 @@ public void put(String path, Object body) {
 		doWrite(createDataPath(path), Collections.singletonMap("data", body));
 	}
 
+	/**
+	 * Performs a KV Patch operation.
+	 * @param path must not be {@literal null} or empty.
+	 * @param kv New key value map to be updated
+	 * @since 2.3
+	 */
+	public void patch(String path, Map<String, ?> kv) {
```
mp911de, line 99: This operation should be also provided for the version 1 backend. Since this class is package-protected, there's no way to consume this method. Therefore, the method should be also made available in the `VaultKeyValueOperations` interface.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
```diff
@@ -89,4 +90,29 @@ public void put(String path, Object body) {
...
+		Assert.hasText(path, "Path must not be empty");
+
+		// To do patch operation, we need to do a read operation first
+		VaultResponse readResponse = get(path);
+		Assert.notNull(readResponse, "VaultResponse must not be null");
```
mp911de, line 104: If the secret doesn't exist, a meaningful exception should be thrown. I think a new `SecretNotFoundException` type (`extends VaultException`) would make sense.

spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueTemplateVersionedIntegrationTests.java
```diff
@@ -35,4 +45,28 @@
...
+		final String newKey = "newKey";
+		Map<String, String> secret = Collections.singletonMap(oldKey, "value");
+
+		String key = UUID.randomUUID().toString();
+
+		this.kvOperations.put(key, secret);
+
+		Map<String, String> newSecret = Collections.singletonMap(newKey, "newValue");
+
+		((VaultKeyValue2Template) this.kvOperations).patch(key, newSecret);
```
mp911de, line 60: Once the `patch` method gets exposed through `VaultKeyValueOperations`, there's no need to have the downcast.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
```diff
@@ -89,4 +90,29 @@ public void put(String path, Object body) {
...
+				String.format("No data found at %s; patch only works on existing data", path));
+		Assert.notNull(readResponse.getMetadata(), "Metadata must not be null");
+		Map<String, Object> data = readResponse.getData();
+		Map<String, Object> metadata = readResponse.getMetadata();
+		kv.forEach(data::put);
+		Map<String, Object> body = new HashMap<>();
+		body.put("data", data);
+		body.put("options", Collections.singletonMap("cas", metadata.get("version")));
+
+		doWrite(createDataPath(path), body);
```
mp911de, line 115: There should be also a reaction on whether the write was successful. A concurrent process could have updated the secret in the meantime. Returning `true` on success (similar to Java's `compareAndSet` methods in the concurrency util package) would make sense.

mp911de (changes requested):
Thanks a lot. I had a look and left a few comments.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
yfaney: Applied:
1. VaultKeyValueOperations.java : https://github.com/spring-projects/spring-vault/pull/587/commits/95d25bd9120c340cecc2eab71c67ffc047e41f07#diff-93f96dc2bd921937c3f5ee0eda355c0dR70
1. VaultKeyValue2Template.java : https://github.com/spring-projects/spring-vault/pull/587/commits/95d25bd9120c340cecc2eab71c67ffc047e41f07#diff-917f198074796659086e92ad34820ddfR101
1. VaultKeyValue1Template.java : https://github.com/spring-projects/spring-vault/pull/587/commits/95d25bd9120c340cecc2eab71c67ffc047e41f07#diff-36bdc62ca9e4f9e53d696f8675e326eaR111

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
yfaney: Applied here: https://github.com/spring-projects/spring-vault/pull/587/commits/95d25bd9120c340cecc2eab71c67ffc047e41f07#diff-917f198074796659086e92ad34820ddfR106-R112

spring-vault-core/src/test/java/org/springframework/vault/core/VaultKeyValueTemplateVersionedIntegrationTests.java
yfaney: Now it does not downcast: https://github.com/spring-projects/spring-vault/pull/587/commits/95d25bd9120c340cecc2eab71c67ffc047e41f07#diff-98a440907d66c701b5a12de4a5db680aR61

spring-vault-core/src/main/java/org/springframework/vault/core/VaultKeyValue2Template.java
yfaney: I changed the return type as boolean and true if the response is successful, but please let me know if you want another return type such as VaultResponse itself or something else:
https://github.com/spring-projects/spring-vault/pull/587/commits/95d25bd9120c340cecc2eab71c67ffc047e41f07#diff-917f198074796659086e92ad34820ddfR120-R125

And FYI, according to the document of HashCorp's KV secret engine V2, we will be safe from the concurrency issue because Vault will not allow if the secret's versioning is not correct.(If the existing version of the secret is 3 and incoming one is still 3 it won't allow writing operation so it can avoid from the race condition)
Please see the following links:
https://learn.hashicorp.com/tutorials/vault/versioned-kv
https://www.vaultproject.io/docs/commands/kv/patch

yfaney:
Hi, @mp911de I addressed all of your comments above. Please review my changes again and let me know if there are still something to be changed. Thanks!

mp911de (approved):
The formatting is a bit off and license headers are missing. That's not a blocker, just wanted to mention it and we will take care of this during the merge.

Going to take this PR from here.

yfaney:
@mp911de awesome, I'll keep that in mind and try following the formatting and license headers in the next time :)

PR #590, opened 2020-09-30
Enabled Vault clients to revoke leases using the revoke-prefix endpoint

### LINKED ISSUE #588: Extend LeaseEndpoints to allow path-based lease revocation
Hey, there!

I'm sorry for bringing this up again, but I feel like #480 was never fully explored or explained, and it's still an annoyance due to false positives in error logs as well as a slight security risk.

I don't believe most admins will grant all their applications access to `sys/leases/revoke`. Doing so would give those applications the ability to revoke any lease as long as they know the lease ID.

With that assumption in mind, I can only see two possible solutions:
- Disable revoking of leases when an application shuts down since it's likely to fail anyway. This only really removes the error messages from the log, but doesn't actually solve the main problem with dangling leases
- Allow applications to revoke a specific lease by their ID via the `sys/leases/revoke-prefix/` endpoint, which is documented [here ](https://www.vaultproject.io/api-docs/system/leases#revoke-prefix)

I know the `revoke-prefix` endpoint also has its caveats, most notably that it'll revoke anything that's "after" whatever prefix you define. I do believe it should be up to each project/organization to decide if they want to leverage it, however. It is possible to give an application access to a specific lease ID while rejecting all other revocation attempts, for example.

One relatively simple (and possibly naïve, I haven't looked through the whole `spring-vault` code base!) solution to the problem could perhaps be to add another constant to the `LeaseEndpoints` enum. That makes it easy for people to opt in by setting the `spring.cloud.vault.config.lifecycle.lease-endpoints` to `SysLeasesPrefix` (or whatever the name ends up being).
```java
// TODO: Come up with a better name, maybe
SysLeasesPrefix {

	@Override
	public void revoke(Lease lease, RestOperations operations) {

		// TODO: Consider if this is secure enough or if it needs to be sanitized
		String endpoint = "sys/leases/revoke-prefix/" + lease.getLeaseId();
		operations.put(endpoint, null);
	}

	@Override
	public Lease renew(Lease lease, RestOperations operations) {

		return SysLeases.renew(lease, operations);
	}
};
```

In any case, keep up the good work! 😃

ThomasKasene:
Changes made to `LeaseEndpoints`:

- Added `LeasesRevokedByPrefix` which uses the `/sys/leases/revoke-prefix/{prefix}` endpoint when revoking leases
- Added `Leases` which work just like `SysLeases`, ie. it uses the `/sys/leases/revoke` endpoint when revoking leases
- Deprecated `SysLeases` in favour of `Leases`
- Did various refactoring to reduce the amount of code duplication

Closes #588.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/lease/LeaseEndpoints.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/lease/LeaseEndpoints.java
@@ -44,13 +44,22 @@ public enum LeaseEndpoints {
 		@Override
 		public void revoke(Lease lease, RestOperations operations) {
 
-			revokeUsing("sys/revoke", lease, operations);
+			operations.exchange("sys/revoke", HttpMethod.PUT, LeaseEndpoints.getLeaseRevocationBody(lease), Map.class,
+					lease.getLeaseId());
 		}
 
 		@Override
+		@SuppressWarnings("unchecked")
 		public Lease renew(Lease lease, RestOperations operations) {
 
-			return renewUsing("sys/renew", lease, operations);
+			HttpEntity<Object> leaseRenewalEntity = getLeaseRenewalBody(lease);
+
+			ResponseEntity<Map<String, Object>> entity = (ResponseEntity) operations.exchange("sys/renew", HttpMethod.PUT,
+					leaseRenewalEntity, Map.class);
+
+			Assert.state(entity != null && entity.getBody() != null, "Renew response must not be null");
+
+			return toLease(entity.getBody());
 		}
 	},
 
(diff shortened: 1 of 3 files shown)


mp911de:
The pull request looks pretty solid. I'd like to keep the implementations as-is meaning not extracting common bits into methods to keep things simple and more readable (and together). It would also make sense to extend the tests, specifically for `SecretLeaseContainerUnitTests` since we're quite lenient in the tests right now.

ThomasKasene:
Hmm, `SecretLeaseContainerUnitTests` mocks out everything at the moment though. Should I change all that or just try to make a `LeaseEndpointsTests` instead?

I can put the code back to its original state, but does that also include the implementation for `SysLeases`, or should that still use the implementation from `Leases` to emphasize that it's the same thing?

By the way, I noticed the argument list for the old revocation implementation includes `lease.getLeaseId()` but that shouldn't be necessary, should it? The lease ID will be passed into the PUT request body:
```java
operations.exchange(endpoint, HttpMethod.PUT, LeaseEndpoints.getLeaseRevocationBody(lease), Map.class,
		lease.getLeaseId());
```

ThomasKasene:
I gambled and added `LeaseEndpointsUnitTests` - it might be useful regardless of whatever should be done about `SecretLeaseContainerUnitTests`, if anything.

While I looked around in `SecretLeaseContainerUnitTests` I took the liberty of replacing the usages of `Mockito.verifyZeroInteractions` with `Mockito.verifyNoInteractions` because the former is deprecated. I still don't know how to improve this test class with regards to the changes made to `LeaseEndpoints`, though.

I also reverted the refactoring in `LeaseEndpoints` and assumed you wanted me to copy the contents from `Leases` into `SysLeases` rather than calling `Leases.renew` and `Leases.revoke`.

mp911de:

PR #595, opened 2020-11-12
Added option to specify certificate format

whiskeysierra:
When issuing certificates or signing CSRs, the format was hard-coded to "der".
This pull request introduces a way to specify a different format, e.g. "pem".


mp911de:
We use DER-encoded certificates to be able to return the certificate as `X509Certificate` or `KeyStore` including the appropriate return type. Changing the format will cause our utility methods to no longer work. That should also work with `pem`-encoded certificates. 

Requesting `pem_bundle` when issuing a certificate should return a `CertificateBundle` instead of just `Certificate` type.

That being said, the PR doesn't fit conceptually the Spring Vault API. If you need to request different certificate types, it would be easier to make `createIssueRequest(…)` public instead.

mp911de:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #596, opened 2020-11-17
Logger is not passed required parameter

igouss:
We should either not log the token and remove %s or log the token.


spring-vault-core/src/main/java/org/springframework/vault/authentication/LifecycleAwareSessionManager.java
```diff
@@ -177,7 +177,7 @@ protected void revoke(VaultToken token) {
 			dispatch(new AfterLoginTokenRevocationEvent(token));
 		}
 		catch (RuntimeException e) {
-			this.logger.warn("Cannot revoke VaultToken: %s", token, e);
+			this.logger.warn("Cannot revoke VaultToken: " + token, e);
```
mp911de, line 180: That should be actually a `String.format(…)`. Care to update the PR and squash all commits into a single one?

PR #599, opened 2020-11-25
Logger is not passed required parameter

igouss:
We should either not log the token and remove %s or log the token.


mp911de:

PR #615, opened 2021-01-14
Added VMSS support to Azure MSI Authentication

### LINKED ISSUE #614: VMSS support for Azure MSI Authentication
The *Vault Plugin: Azure Auth Backend* has support for both, standalone VMs and VM scale sets (VMSS):

> * `vm_name (string: "")` - The virtual machine name for the machine that generated the MSI token. This information can be obtained through instance metadata. If vmss_name is provided, this value is ignored.
> * `vmss_name (string: "")` - The virtual machine scale set name for the machine that generated the MSI token. This information can be obtained through instance metadata.
>
> https://www.vaultproject.io/api-docs/auth/azure#login

Spring Vault only supports standalone VMs right now, i.e. it will always set the `vm_name` property when sending a login request.

The `/metadata/instance` request will return a scale set name, if the VM is part of a scale set:

```yaml
{
  "name": "myScaleSet_0",
  "vmScaleSetName": "myScaleSet",
  # ..
}
```

Standalone VMs, which are not part of a scale set, have an empty scale set name:

```yaml
{
  "name": "myVm",
  "vmScaleSetName": "",
  # ..
}
```

## Implementation notes

- `AzureVmEnvironment` should hold an additional property: `String vmssName`
- `AzureMsiAuthentication#toAzureVmEnvironment` should:
  - read `vmScaleSetName`
  - set `vm_name` if `vmScaleSetName` is empty
  - set `vmss_name` otherwise

whiskeysierra:
Fixes #614

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/AzureVmEnvironment.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/AzureVmEnvironment.java
@@ -37,6 +37,8 @@ public class AzureVmEnvironment {
 
 	private final String vmName;
 
+	private final String vmScaleSetName;
+
 	/**
 	 * Creates a new {@link AzureVmEnvironment}.
 	 * @param subscriptionId must not be {@literal null}.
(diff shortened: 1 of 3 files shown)


spring-vault-core/src/main/java/org/springframework/vault/authentication/AzureVmEnvironment.java
```diff
@@ -37,21 +37,26 @@
...
+
 	/**
 	 * Creates a new {@link AzureVmEnvironment}.
 	 * @param subscriptionId must not be {@literal null}.
 	 * @param resourceGroupName must not be {@literal null}.
 	 * @param vmName must not be {@literal null}.
+	 * @param vmScaleSetName must not be {@literal null}.
 	 */
-	public AzureVmEnvironment(String subscriptionId, String resourceGroupName, String vmName) {
+	public AzureVmEnvironment(String subscriptionId, String resourceGroupName, String vmName, String vmScaleSetName) {
```
mp911de, line 49: Care to add another constructor to retain binary compatibility? We should have both constructors, one with and one without `vmScaleSet`.

mp911de (changes requested):

spring-vault-core/src/main/java/org/springframework/vault/authentication/AzureVmEnvironment.java
whiskeysierra: Done

mp911de:

Issue #635, opened 2021-03-09
Add ability to configure client TLS enabled protocol versions and cipher suites via Spring properties


mryangza:
It would be useful if the enabled TLS protocol versions and cipher suites used by an HTTP client to Vault could be configurable via Spring properties (bound to the `org.springframework.vault.support.SslConfiguration` class). These could then be set on the SSL socket factory created by the underlying request factories created via the `org.springframework.vault.client.ClientHttpRequestFactoryFactory` class. It looks like Apache HTTP Components and Netty support setting this via their builders, I am unsure whether OkHttp does. It might require a custom SSL socket factory wrapper that decorates the sockets as they are created (before they are connected).

Thoughts? I am happy to attempt a PR.

mp911de:
Sounds good. We have a config class for ssl that we use to configure the ssl context. Feel free to submit a pull request.

mryangza:
Great, I've already spent some time implementing this, I'll try and finish it off in the next little while.

PR #640, opened 2021-03-15
Add ability to configure client TLS enabled protocol versions and cipher suites via Spring properties

mryangza:
- Adding the ability to explicitly configure the enabled SSL protocol
versions and cipher suites used by the Vault HTTP client via the
following Spring properties:

  * vault.ssl.enabled-protocols
  * vault.ssl.enabled-cipher-suites

- Properties should be a comma-separated list of String constants that
correspond to those used by the enabled SSL provider.

Closes gh-635


mp911de:

PR #644, opened 2021-03-17
Bumped Azure IMDS API version

whiskeysierra:
vmScaleSetName is only exposed in 2017-12-01 and later.
See https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service\?tabs\=windows\#response-1


mp911de:
Thank you for your contribution. That's merged and backported now.

Issue #566, opened 2020-06-12
Support for LDAP authentication 


kuldikin:
Hello, does it make sense to support LDAP authentication? If so, i can do it.

mp911de:
Typically, applications use authentication schemes that qualify for machine authentication. LDAP/username-based schemes are human schemes and therefore we don’t support these. Each time there’s a username/password combination used for machine authentication it indicates a improper authentication scheme.

kuldikin:
Thanks for the answer.  At my company, the admins did not want to support two authentication methods, and we were forced to make an ldap implementation.  I agree that this is not the right approach and it is not worth promoting it to the masses.  Ok, I'm closing the ticket.

PR #668, opened 2021-11-03
Provide userpass and ldap support

sokomishalov:
@mp911de I do agree with your concerns [here](https://github.com/spring-projects/spring-vault/issues/566) about human-machine and machine-machine authentication. But despite that, the truth is some enterprise companies rely too much on LDAP authentication, unfortunately. That forces users to build their own workarounds over that restrictions to achieve some goal.
For example, [teamcity plugin for hashicorp vault](https://github.com/JetBrains/teamcity-hashicorp-vault-plugin/issues/45) which cannot provide `ldap`/`userpass` auth because of absence of implementation in spring-vault which is used by that plugin. This is the case of human-machine-machine authentication so I suppose it's ok to deal with human credentials.
I propose to merge these changes which implement `userpass` and `ldap` auth backends but mark it somehow as unrecommended for usage (just like`@OptIn` from kotlin).

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/LdapAuthentication.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/LdapAuthentication.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.vault.authentication;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.util.Assert;
+import org.springframework.vault.VaultException;
+import org.springframework.vault.client.VaultResponses;
+import org.springframework.vault.support.VaultResponse;
+import org.springframework.vault.support.VaultToken;
+import org.springframework.web.client.HttpStatusCodeException;
+import org.springframework.web.client.RestOperations;
+
+import static java.util.Collections.singletonMap;
+import static org.springframework.vault.authentication.AuthenticationUtil.getLoginPath;
+
+/**
+ * LDAP implementation of {@link ClientAuthentication}.
+ *
+ * @author Mikhael Sokolov
+ * @see LdapAuthenticationOptions
+ * @see RestOperations
+ * @see <a href="https://www.vaultproject.io/docs/auth/ldap">LDAP</a>
+ * @since 2.4
+ */
+public class LdapAuthentication implements ClientAuthentication, AuthenticationStepsFactory {
+
+    private static final Log logger = LogFactory.getLog(LdapAuthentication.class);
+
+    private final LdapAuthenticationOptions options;
+
+    private final RestOperations restOperations;
+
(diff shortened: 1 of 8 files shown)


spring-vault-core/src/main/java/org/springframework/vault/authentication/LdapAuthenticationOptions.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class LdapAuthenticationOptions extends UsernamePasswordAuthenticationOptions {
+
+    private LdapAuthenticationOptions(String username, String password) {
+        super(username, password);
+    }
+
+    @Override
+    public String getPath() {
+        return String.format("auth/ldap/login/%s", username);
```
mp911de, line 29: Let's configure the path through options.

spring-vault-core/src/main/java/org/springframework/vault/authentication/UsernamePasswordAuthentication.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Username and password implementation of {@link ClientAuthentication}.
+ *
+ * @author Mikhael Sokolov
+ * @see UsernamePasswordAuthenticationOptions
+ * @see RestOperations
+ * @see <a href="https://www.vaultproject.io/docs/auth/userpass">Username & password</a>
+ * @since 2.4
+ */
+public class UsernamePasswordAuthentication implements ClientAuthentication {
```
mp911de, line 39: It would make sense to also add an authentication steps implementation to support reactive authentication.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LdapAuthentication.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * LDAP implementation of {@link ClientAuthentication}.
+ *
+ * @author Mikhael Sokolov
+ * @see LdapAuthenticationOptions
+ * @see RestOperations
+ * @see <a href="https://www.vaultproject.io/docs/auth/ldap">LDAP</a>
+ * @since 2.4
+ */
+public class LdapAuthentication extends UsernamePasswordAuthentication {
```
mp911de, line 29: Let's not subclass `UsernamePasswordAuthentication`. While username/password are the primary driver for LDAP, future implementations could bring alternative schemes for LDAP and we should not constrain ourselves.

mp911de (changes requested):
Point taken. So let's add both auth schemes. Besides the existing comments, please switch each occurrence of `password` from `String` to `CharSequence` to allow usage of alternative string representations such as `StringBuilder` as `String` is aggressively cached by the JVM.

sokomishalov:
@mp911de thanks! I've reworked it according to your review.

spring-vault-core/src/main/java/org/springframework/vault/authentication/LdapAuthenticationOptions.java
sokomishalov: done

spring-vault-core/src/main/java/org/springframework/vault/authentication/UsernamePasswordAuthentication.java
sokomishalov: implemented for reactive as well

spring-vault-core/src/main/java/org/springframework/vault/authentication/LdapAuthentication.java
sokomishalov: I agree, fixed

mp911de:
Given that the Login API for both authentication mechanisms looks pretty much the same, I tend to remove the LDAP variant and suggest using `UsernamePasswordAuthentication` for `userpass` and `ldap`. Furthermore, Okta and RADIUS follow the same login scheme. Okta supports additionally one-time tokens (`totp`). Introducing `totp` to `UsernamePasswordAuthenticationOptions` is likely less invasive than having options for each individual authentication mechanism.

I'm going to adopt the PR into that direction.

mp911de:

Issue #676, opened 2022-01-02
Missing "private_key_type" property in VaultCertificateResponse


abremora:
# Description

The `VaultCertificateResponse`  doesn't contain `private_key_type` property.

See: https://www.vaultproject.io/api-docs/secret/pki#generate-certificate

# Solution
Extend `CertificateBundle`

# Environment
spring-vault v2.3.2

abremora:
Thank you for your help. Planned for v2.4, right? Can you please update the milestone tag.

PR #677, opened 2022-01-03
Add private_key_type property in VaultCertificateResponse

abremora:
Fixes gh-676


abremora:
I added some overloads to be compatible to older versions and marked old methoads as deprecated.

spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
```diff
@@ -46,14 +46,43 @@
...
+	 * Create a new {@link CertificateBundle}.
+	 * @param serialNumber the serial number.
+	 * @param certificate the certificate.
+	 * @param issuingCaCertificate the issuing CA certificate.
+	 * @param caChain the CA chain.
+	 * @param privateKey the private key.
+	 * @deprecated since 2.3.3, use {@link #CertificateBundle(String, String, String, List, String, String)} instead.
+	 */
+	@Deprecated
+	CertificateBundle(String serialNumber, String certificate, String issuingCaCertificate,
```
mp911de, line 63: We do not need to reintroduce a deprecated constructor variant as the original constructor was package-private so we're free to change the constructors.

spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
```diff
@@ -78,13 +110,43 @@ public static CertificateBundle of(String serialNumber, String certificate, Stri
...
 	 * @return the private key (decrypted form, PEM or DER-encoded)
 	 */
 	public String getPrivateKey() {
 		return this.privateKey;
 	}
 
+	/**
+	 * @return the private key type.
+	 */
+	public String getPrivateKeyType() {
```
mp911de, line 146: Nit: this needs to be `@Nullable`

mp911de (approved):
Thanks for submitting a pull request. I'll have a look throughout the next days.

abremora:
@mp911de : Sorry for the delay, I would have continued to work on the PR throughout the day. You already fixed it and pushed it to main branch?

mp911de:
No worries, yeah, that's already on `main`.

PR #686, opened 2022-02-04
Replace algorithm property in VaultSignRequest/VaultSignatureVerificationRequest

Niamsy:
Based on [https://www.vaultproject.io/api/secret/transit#sign-data](url) and [https://www.vaultproject.io/api/secret/transit#verify-signed-data](url), algorithm property does no longer exist and was replaced by hash_algorithm. When using a RSA key signature_algorithm is also used.

Replace algorithm property with hashAlgorithm and signatureAlgorithm in VaultSignRequest and VaultSignatureVerificationRequest and update mapping in `VaultTransitTemplate.sign()` and `VaultTransitTemplate.verify()`

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultTransitTemplate.java
@@ -384,8 +384,12 @@ public class VaultTransitTemplate implements VaultTransitOperations {
 		Map<String, Object> request = new LinkedHashMap<>();
 		request.put("input", Base64Utils.encodeToString(signRequest.getPlaintext().getPlaintext()));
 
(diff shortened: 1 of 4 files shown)


mp911de:
When were these properties introduced and when was the `algorithm` property removed? We need to carefully consider the version range to ensure what Vault versions we want to support.

Niamsy:
I coudn't find this information in any of the release notes on vault website, but after some research in vault repository, `hash_algorithm` and `signature_algorithm` seem to have been introduced in v0.9.6 [https://github.com/hashicorp/vault/blob/v0.9.6/builtin/logical/transit/path_sign_verify.go](url) and `algorithm` has been noted deprecated since then. 

v0.9.5 [https://github.com/hashicorp/vault/blob/v0.9.5/builtin/logical/transit/path_sign_verify.go](url) still uses `algorithm`

Issue #683, opened 2022-01-05
Support elliptic curve keys in `CertificateBundle` 


abremora:
# Description

No keystore can be created when the private key is of type `private_key_type=ec`.

```
java.lang.IllegalStateException: Invalid DER: object (4) is not integer.
	at org.springframework.vault.support.KeystoreUtil$Asn1Object.getInteger(KeystoreUtil.java:502) ~[spring-vault-core-2.3.2.jar:2.3.2]
	at org.springframework.vault.support.KeystoreUtil.getRSAPrivateKeySpec(KeystoreUtil.java:255) ~[spring-vault-core-2.3.2.jar:2.3.2]
	at org.springframework.vault.support.CertificateBundle.getPrivateKeySpec(CertificateBundle.java:90) ~[spring-vault-core-2.3.2.jar:2.3.2]
	at org.springframework.vault.support.CertificateBundle.createKeyStore(CertificateBundle.java:109) ~[spring-vault-core-2.3.2.jar:2.3.2]
````

# Solution
Fix `CertificateBundle` and its dependencies like ` KeystoreUtil` which are focused on RSA implementation.

abremora:
I have a working version but I use Bouncy Castle for EC support. Is there anything we need to consider when using third party libs?

mp911de:
`KeystoreUtil` is a core utility that should remain free of 3rd-party dependencies.

abremora:
I didn't use the `KeystoreUtil` for EC implementation. All details are implemented in a separate class. Currently Bouncy Castle is an optional dependency in spring-vault, so we have to include BC only when EC should be used. Is this an option?

mp911de:
I think so. We can guard the access to BC and document that requirement. But it's not something that is obvious. Ideally, we would have an EC-capable certificate/key parser in our code.

abremora:
However, there is already an existing implementation in BC for our problem and it is quite easy to use:
```java
private static java.security.spec.ECPrivateKeySpec getEcPrivateKeySpec(byte[] privateKey) {
    ECPrivateKey ecPrivateKey = ECPrivateKey.getInstance(privateKey);
    ASN1ObjectIdentifier parameters = (ASN1ObjectIdentifier) ecPrivateKey.getParameters();

    X9ECParameters curveParameter = X962NamedCurves.getByOID(parameters);

    EllipticCurve ellipticCurve = new EllipticCurve(
[...]

Issue #678, opened 2022-01-04
Add support to obtain PEM-encoded certificates as `X509Certificate`


abremora:
`Certificate.getX509Certificate()` and `Certificate.getX509IssuerCertificate()` do not support conversion of PEM-encoded string to X509Certificate.

abremora:
It's odd in that the default of the returned data is PEM.

> format (string: "") – Specifies the format for returned data. Can be pem, der, or pem_bundle; defaults to pem. 
> If der, the output is base64 encoded. If pem_bundle, the certificate field will contain the private key and certificate, 
> concatenated; if the issuing CA is not a Vault-derived self-signed root, this will be included as well.

https://www.vaultproject.io/api-docs/secret/pki#generate-certificate

abremora:
Same issue affects  `CertificateBundle.createKeyStore()` & `CertificateBundle.getPrivateKeySpec()`, too. (Already mentioned in the documentation)

```
java.lang.IllegalArgumentException: Illegal base64 character 2d
	at java.base/java.util.Base64$Decoder.decode0(Base64.java:746) ~[na:na]
	at java.base/java.util.Base64$Decoder.decode(Base64.java:538) ~[na:na]
	at org.springframework.util.Base64Utils.decode(Base64Utils.java:59) ~[spring-core-5.3.13.jar:5.3.13]
	at org.springframework.util.Base64Utils.decodeFromString(Base64Utils.java:111) ~[spring-core-5.3.13.jar:5.3.13]
	at org.springframework.vault.support.CertificateBundle.getPrivateKeySpec(CertificateBundle.java:89) ~[spring-vault-core-2.3.2.jar:2.3.2]
	at org.springframework.vault.support.CertificateBundle.createKeyStore(CertificateBundle.java:109) ~[spring-vault-core-2.3.2.jar:2.3.2]
````

abremora:
Just for clarification: For `Vault` the default is PEM but for `spring-vault` the default is DER:

https://github.com/spring-projects/spring-vault/blob/26a77a2199ecd5186f04c930c2062ef7bb4711ee/spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiTemplate.java#L91

I will fix this issue within the next days.

abremora:
Of course we cannot change the behavior of vault-spring but we have to support other formats and improve documentation.

PR #688, opened 2022-03-07
Support PEM encoded certificates and EC for private keys

abremora:
Support for PEM encoded certificates added. Although there is a parser for PEM files already available (see `PemObject`), I introduced a new one called `PemReader`. This implementation has no dependencies to RSA. This seems to be reasonable because there are more algorithms available. Further I looked for a solution to add easily more private key implementations. The result was a `PrivateKeyFactory` which holds a list of private key strategies (`RsaPrivateKeyStrategy`, `EcPrivateKeyStrategy`).
`RsaPrivateKeyStrategy` is just a wrapper for `KeystoreUtil.getRSAPrivateKeySpec()`. 
To prove the concept, I quickly develop a `EcPrivateKeyStrategy`. That's why the support for EC is already present here. If you like to split these two features into separate PRs just give me a note.

Fixes gh-678
Fixes gh-683


spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
```diff
@@ -39,103 +38,121 @@
...
+
+  private final CertificateEncoding issuingCaCertificateFormat;
+
+  Certificate(@JsonProperty("serial_number") String serialNumber, @JsonProperty("certificate") String certificate,
+      @JsonProperty("issuing_ca") String issuingCaCertificate) {
+
+    this.serialNumber = serialNumber;
+    this.certificate = certificate;
+    this.issuingCaCertificate = issuingCaCertificate;
+    this.certificateFormat = CertificateEncodingValidator.getFormat(certificate);
```
abremora, line 57: We have no information about the current encoding so we have to calculate it for further processing. Should this an enum or a string only?

spring-vault-core/src/main/java/org/springframework/vault/support/EcPrivateKeyStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+package org.springframework.vault.support;
+
+import java.security.spec.ECFieldFp;
+import java.security.spec.ECParameterSpec;
+import java.security.spec.ECPoint;
+import java.security.spec.ECPrivateKeySpec;
+import java.security.spec.EllipticCurve;
+import java.security.spec.KeySpec;
+import org.bouncycastle.asn1.ASN1ObjectIdentifier;
```
abremora, line 24: We use Bouncy Castle here for EC support. Can we use this dependency here?

mp911de: Ideally, we don't use BouncyCastle at all as the cost for the dependency is quite large in comparison to what we get. We have a DER parser already and it's a matter of parsing the ASN.1 sequence. Java's crypto utils contain proper support to resolve EC parameters through an OID and provide key specs.

I have the code ready for this.

abremora:
I am happy to support you. Please do not hesitate to contact me if you need further assistance.

Issue #708, opened 2022-05-30
Create keystore with password


abremora:
No password will be set by `CertificateBundle.createKeyStore()`. There should be a new overload which provides a parameter to set the password. 

https://github.com/spring-projects/spring-vault/blob/dd9c3cfab8aca6ee4572f9c81b9a02000b2db480/spring-vault-core/src/main/java/org/springframework/vault/support/KeystoreUtil.java#L94

mp911de:
Good suggestion. Feel free to submit a pull request.

sodrac:
Hi @mp911de, if this issue is still available I'll like to work on it.

mp911de:
Sure, feel free to submit a pull request.

mp911de:
Thank you for your contribution. That's merged, polished, and backported now.

PR #711, opened 2022-06-15
Overload createKeyStore to allow password parameter to be set Fixes gh-708

sodrac:
- Overload the CertificateBundle.createKeyStore() to allow a String password parameter to be set
- Overload the static KeystoreUtil.createKeyStore() to allow a String password parameter to be set
- Added javadoc and unit tests
- Code formatting, best practices and cosmetic changes
- Fixes gh-708


spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
```diff
@@ -214,6 +227,41 @@ public KeyStore createKeyStore(String keyAlias, boolean includeCaChain) {
...
+	 * Create a {@link KeyStore} from this {@link CertificateBundle} containing the
+	 * private key and certificate chain.
+	 * @param keyAlias the key alias to use.
+	 * @param includeCaChain whether to include the certificate authority chain instead of
+	 * just the issuer certificate.
+	 * @param password the password to use.
+	 * @return the {@link KeyStore} containing the private key and certificate chain.
+	 * @since 3.0.0
+	 */
+	public KeyStore createKeyStore(String keyAlias, boolean includeCaChain, String password) {
```
mp911de, line 240: It makes sense to accept `CharSequence` instead of `String` as password (and have another overload accepting `char[]`) as `String`s are heavily cached by the JVM. Also, the new method looks 90% similar to the already existing one. There's room to eliminate duplicate code by having an implementation that can handle absence/presence of the password.

sodrac: changes are committed and ready for review.

mp911de:

PR #757, opened 2023-02-22
Added getting support links

esivakumar18:


mp911de:
Thanks for your suggestion. There's already a note in the  `Getting Help` section. Also, we do not want to advertise commercial support in our readme.

Issue #758, opened 2023-02-24
AWS IAM Authentication should provide a way to sign request for the global STS endpoint (eu-east-1)


Amuerte:
Since `v3.0.x`, the `AwsIamAuthentication` is using the AWS region computed by `software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain`, especially to sign the headers for the AWS v4 API, _which is a good point_. 

But in some cases, the Vault instance which is making the proxy call to AWS, is configured to use the **_global_** STS endpoint (on `us-east-1`), which is the [default behavior](https://github.com/hashicorp/vault/blob/main/builtin/credential/aws/path_login.go#L292). And for security reasons, Vault is [not using](https://github.com/hashicorp/vault/blob/main/builtin/credential/aws/path_login.go#L1707) the `iam_request_url` parameter.

As a consequence, you end up with an application deployed in a region A and a Vault instance using the global STS endpoint, without no easy way to use the `spring-vault` client to login.

Therefore, I think we should provide a way to sign request for the global STS endpoint (eu-east-1) by offering a dedicated option and 
I will be glad to help on that.

mp911de:
I'm not too familiar with the AWS API nor the SDK, we would be happy if you would submit a pull request so we can discuss the actual change and make it work in the context of Spring Vault.

Amuerte:
Changes done in PR #763

Amuerte:
Hi @mp911de :wave: 

Any chances to see this shipped with release `3.0.2` ?

mp911de:
Looking at the PR, we already provide a way to set the AWS region. Having a method named `useGlobalRegion` that sets the region doesn't make sense as the region can be set already.

PR #763, opened 2023-03-04
Support authentication on global region for AWS IAM

Amuerte:
This PR enables one to perform AWS IAM authentication when the Vault instance is configured on the Global STS endpoint (us-east-1) and the application is deployed on another region.


Amuerte:
Close #758

spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthenticationOptions.java
```diff
@@ -282,16 +294,27 @@ public AwsIamAuthenticationOptionsBuilder endpointUri(URI endpointUri) {
...
+		}
+
 		/**
 		 * Build a new {@link AwsIamAuthenticationOptions} instance.
 		 * @return a new {@link AwsIamAuthenticationOptions}.
 		 */
 		public AwsIamAuthenticationOptions build() {
 
 			Assert.state(this.credentialsProvider != null, "Credentials or CredentialProvider must not be null");
+			if (useGlobalEndpoint) {
```
mp911de, line 312: If using the global endpoint is merely a matter of providing the right `Region`, why don't you just call `regionProvider(() -> Region.US_EAST_1)`?

mp911de (changes requested):
I'm not quite sure this pull request makes sense as all we do is setting a `Region`

spring-vault-core/src/main/java/org/springframework/vault/authentication/AwsIamAuthenticationOptions.java
Amuerte: mmh good point you raised here :sweat_smile: . By looking more in details, we are blocked because we are using [spring-cloud-vault-config](https://github.com/spring-cloud/spring-cloud-vault/blob/3.1.x/spring-cloud-vault-config/src/main/java/org/springframework/cloud/vault/config/ClientAuthenticationFactory.java#L276) to instantiate the AwsIamAuthentication, and it does not offer a way to configure the region.
I will create an issue and a PR there. Sorry for that.

PR #780, opened 2023-04-13
Role name can now be used with cert auth

andylintner:
When using TLS authentication, the Vault API allows the caller to specify a role to use for authorization. Previously, we did not allow this to be specified, which caused Vault to select one of the roles associated with the certificate.


mp911de:
Thanks for reaching out. We handle backports ourselves. Generally, we carefully consider which features we want to port back into older versions. This one comes with little risk so we can ship this feature within a 2.3.x release.

mp911de:

andylintner:
Thanks @mp911de! What's the release cadence like for this project? Just wondering when I can expect to be able to consume these changes.

andylintner:
@mp911de - any chance of getting a 2.3.4 release cut?

Issue #745, opened 2022-12-09
Support `prehashed` for the Transit engine


Gentoli:
https://developer.hashicorp.com/vault/api-docs/secret/transit#prehashed

>[prehashed](https://developer.hashicorp.com/vault/api-docs/secret/transit#prehashed) (bool: false) - Set to true when the input is already hashed. If the key type is rsa-2048, rsa-3072 or rsa-4096, then the algorithm used to hash the input should be indicated by the hash_algorithm parameter. Just as the value to sign should be the base64-encoded representation of the exact binary data you want signed, when set, input is expected to be base64-encoded binary hashed data, not hex-formatted. (As an example, on the command line, you could generate a suitable input via openssl dgst -sha256 -binary | base64.)

mp911de:
Feel free to submit a pull request.

PR #791, opened 2023-05-28
Support prehashed for the Transit engine

nbaars:
Close gh-745

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/support/Plaintext.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/support/Plaintext.java
@@ -17,8 +17,8 @@ package org.springframework.vault.support;
 
 import java.nio.charset.Charset;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.Objects;
-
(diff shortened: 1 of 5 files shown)


spring-vault-core/src/main/java/org/springframework/vault/support/Plaintext.java
```diff
@@ -138,6 +138,19 @@ public String asString(Charset charset) {
 		return new String(getPlaintext(), charset);
 	}
 
+	/**
+	 * @return <tt>true</tt> if the plaintext is base64 encoded
+	 */
+	public boolean isBase64() {
```
mp911de, line 144: I'm not sure this is something we should check. by decoding. The payload can be rather big and decoding it on our side imposes quite some load. If the payload isn't Base64-encoded, then Vault will report the issue in the HTTP response.

nbaars: @mp911de correct, I did not think about the size of the possible plaintext and indeed Vault will report an error. I've updated the PR

mp911de:

Issue #689, opened 2022-03-18
Support JWT Authentication


paguos:
Add the JWT authentication method to the library: [https://www.vaultproject.io/api-docs/auth/jwt](https://www.vaultproject.io/api-docs/auth/jwt)

paguos:
Is this something you could consider? 😃

mp911de:
Feel free to submit a pull request including ideally some integration tests that we can run in an isolated environment (i.e. without internet/external network access).

PR #802, opened 2023-06-28
Support JWT Authentication

nbaars:
Close gh-689

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/authentication/JwtAuthenticationOptions.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/authentication/JwtAuthenticationOptions.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2017-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.vault.authentication;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+import org.springframework.util.Assert;
+
+/**
+ * Authentication options for {@link JwtAuthentication}.
+ * <p>
+ * Authentication options provide the role and the JWT. {@link JwtAuthenticationOptions}
+ * can be constructed using {@link #builder()}. Instances of this class are immutable once
+ * constructed.
+ * <p>
+ *
+ * @author Nanne Baars
+ * @since 3.0.4
+ * @see JwtAuthentication
+ * @see #builder()
+ */
+public class JwtAuthenticationOptions {
+
+	/**
+	 * Path of the JWT authentication backend mount. Optional and defaults to
+	 * {@literal jwt}.
+	 */
+	private final Optional<String> path;
+
+	/**
+	 * Name of the role against which the login is being attempted. Defaults to configured
+	 * default_role if not provided. See
+	 * <a href="https://developer.hashicorp.com/vault/api-docs/auth/jwt#configure">Vault
+	 * JWT configuration</a>
+	 */
+	private final Optional<String> role;
+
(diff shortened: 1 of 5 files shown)


spring-vault-core/src/main/java/org/springframework/vault/authentication/JwtAuthenticationOptions.java
```diff
@@ -0,0 +1,156 @@
...
+	 * default_role if not provided. See
+	 * <a href="https://developer.hashicorp.com/vault/api-docs/auth/jwt#configure">Vault
+	 * JWT configuration</a>
+	 */
+	private final Optional<String> role;
+
+	/**
+	 * The signed JSON Web Token (JWT) to be used for authentication.
+	 */
+	private final String jwt;
```
mp911de, line 54: This should be a supplier to allow integrating with JWT's obtained from somewhere else, which can expire over time.

spring-vault-core/src/test/java/org/springframework/vault/authentication/JwtAuthenticationIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.http.MediaType;
+import org.springframework.vault.util.IntegrationTestSupport;
+import org.springframework.vault.util.Settings;
+import org.springframework.vault.util.TestRestTemplateFactory;
+
+public class JwtAuthenticationIntegrationTest extends IntegrationTestSupport {
```
mp911de, line 49: Thanks for the elaborate integration test. It might be easier to use a mocked setup (see `GcpIamAuthenticationUnitTests`) as there's no computation involved for passing on the JWT token. Having another webserver introduces complexity in the form of Vault configuration and another moving part.

src/main/asciidoc/reference/authentication.adoc
```diff
@@ -722,6 +722,8 @@ class AppConfig extends AbstractVaultConfiguration {
 
     // …
 }
+
```
mp911de, line 725: Seems some changes made it into the documentation that shouldn't be here.

spring-vault-core/src/main/java/org/springframework/vault/authentication/JwtAuthenticationOptions.java
nbaars: Ah I had this in the first implementation I will update this

spring-vault-core/src/test/java/org/springframework/vault/authentication/JwtAuthenticationIntegrationTest.java
nbaars: Tested it first with a local Keycloak server to make sure it worked. Then moved to Keycloak mock library which had a bug. 

But you are right the moving components are not necessary I will change it to a more fixed setup

spring-vault-core/src/main/java/org/springframework/vault/authentication/JwtAuthenticationOptions.java
nbaars: Updated and moved to supplier

spring-vault-core/src/main/java/org/springframework/vault/authentication/JwtAuthenticationOptions.java
```diff
@@ -0,0 +1,156 @@
...
+ * @see JwtAuthentication
+ * @see #builder()
+ */
+public class JwtAuthenticationOptions {
+
+	/**
+	 * Path of the JWT authentication backend mount. Optional and defaults to
+	 * {@literal jwt}.
+	 */
+	private final Optional<String> path;
```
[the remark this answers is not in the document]
nbaars: Updated

spring-vault-core/src/test/java/org/springframework/vault/authentication/JwtAuthenticationIntegrationTest.java
nbaars: Moved to the PEM variant for JWT authentication which prevents the extra dependency on the mock-server completely.

src/main/asciidoc/reference/authentication.adoc
nbaars: Indeed, c&p mistake

mp911de:
Thanks a lot for revisiting this PR. I'm happy to take this pull request from here.

Issue #690, opened 2022-03-28
Support role_name and entity_alias in VaultTokenRequest


edmundham:
Hi all,

From the vault documentation for creating token through API call, https://www.vaultproject.io/api-docs/auth/token#create-token they have `role_name` and `entity_alias`. However, current (2.3.2 version) `VaultTokenRequest` does not support these parameters. https://github.com/spring-projects/spring-vault/blob/2.3.2/spring-vault-core/src/main/java/org/springframework/vault/support/VaultTokenRequest.java

Can the team look into adding these?

Thanks in advance!

mp911de:
Feel free to submit a pull request.

PR #803, opened 2023-07-02
Support role_name and entity_alias in VaultTokenRequest

nbaars:
Closes gh-690


mp911de:

PR #807, opened 2023-07-20
Draft: ReactiveVaultTemplate for the key-value backend version 2

tweiand:
What is done:
- Add version 2 (by porting blocking code APIs)
- Unit tests for all additional classes
- Maybe some code deduplication and cleanup
- A second round of editing java docs

Questions I have:
- Is author attribution correct since this can be considered a port of the block code?
- What do you think about the change?


spring-vault-core/pom.xml
```diff
@@ -324,4 +324,89 @@
 		</dependency>
 	</dependencies>
 
+	<profiles>
```
mp911de, line 327: We do not use Sonar so please remove the additional profile.

spring-vault-core/src/main/java/org/springframework/vault/core/ReactiveVaultKeyValue2Template.java
```diff
@@ -0,0 +1,147 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.vault.core;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import org.eclipse.jetty.util.ssl.SslContextFactory.Client;
```
mp911de, line 22: The Jetty import seems not to be correct.

spring-vault-core/src/main/java/org/springframework/vault/core/ReactiveVaultKeyValueMetadataTemplate.java
```diff
@@ -0,0 +1,128 @@
...
+	ReactiveVaultKeyValueMetadataTemplate(ReactiveVaultOperations vaultOperations, String basePath) {
+
+		Assert.notNull(vaultOperations, "VaultOperations must not be null");
+
+		this.vaultOperations = vaultOperations;
+		this.basePath = basePath;
+	}
+
+	@SuppressWarnings({ "ConstantConditions", "unchecked", "rawtypes" })
+	private static VaultMetadataResponse fromMap(Map<String, Object> metadataResponse) {
```
mp911de, line 54: We should avoid code duplications and move utility methods into a better place.

spring-vault-core/src/main/java/org/springframework/vault/support/VaultResponseDataVersion2.java
```diff
@@ -0,0 +1,34 @@
+package org.springframework.vault.support;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.util.Map;
+import org.springframework.lang.Nullable;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class VaultResponseDataVersion2<T> {
```
mp911de, line 8: Any reason why we require an additional response wrapper and why we cannot follow the imperative typing approach?

spring-vault-core/src/test/java/org/springframework/vault/core/ReactiveVaultKeyValueMetadataTemplateIntegrationTests.java
```diff
@@ -0,0 +1,156 @@
...
+
+		vaultKeyValueMetadataOperations.put(CAS_SECRET_NAME, request).as(StepVerifier::create).verifyComplete();
+
+		vaultKeyValueMetadataOperations.get(CAS_SECRET_NAME)
+			.as(StepVerifier::create)
+			.assertNext(metadataResponseAfterUpdate -> {
+				assertThat(metadataResponseAfterUpdate.isCasRequired()).isEqualTo(request.isCasRequired());
+				assertThat(metadataResponseAfterUpdate.getMaxVersions()).isEqualTo(request.getMaxVersions());
+
+				if (prepare().getVersion().isGreaterThanOrEqualTo(Version.parse("1.2.0"))) {
```
mp911de, line 134: Nit: We should obtain the version outside of the reactive flow to avoid potential blocking issues.

spring-vault-core/src/main/java/org/springframework/vault/core/ReactiveKeyValueHelper.java
```diff
@@ -0,0 +1,36 @@
...
+package org.springframework.vault.core;
+
+import org.springframework.vault.support.VaultResponseSupport;
+import reactor.core.publisher.Mono;
+
+/**
+ * Helper for wrapping imperative operations.
+ *
+ * @author Timothy R. Weiand
+ * @since 3.999.999
```
mp911de, line 25: The target version will be `3.1`, so `@since 3.1` is the way to go across all newly introduced classes and methods.

mp911de (changes requested):
I had a first look. The structure looks pretty much like the imperative support. There are a few deviations that we should discuss, I left a few comments.

spring-vault-core/src/main/java/org/springframework/vault/core/ReactiveVaultKeyValue2Template.java
tweiand: Great catch!

spring-vault-core/src/main/java/org/springframework/vault/support/VaultResponseDataVersion2.java
tweiand: My goal was to force WebClient to do all JSON marshaling. My intent was to use any added Jackson deserialization to the DI'ed ` WebClient.Builder`.

To do what imperative did - do you know of a way to get the default Jackson deserialization configuration from WebClient?

spring-vault-core/src/test/java/org/springframework/vault/core/ReactiveVaultKeyValueMetadataTemplateIntegrationTests.java
tweiand: Will update.

tweiand:
@mp911de - I had a bad git config when doing most of the work on this. After we finish review, would it be possible for me to re-make the branch so that the commits look correct?

mp911de:
Sure. A force-push is fine.
Am 26. Aug. 2023, 01:10 +0200 schrieb Timothy R. Weiand ***@***.***>:

tweiand:
Force pushed the branch with the correct email for me. I have 2 questions buried in the comments. Thanks for the help!

Issue #687, opened 2022-02-28
Add transit rewrap batch method


ShuP1:
Just a feature request to match Transit API in a better way

Vault provides `batch_input` on `transit/%s/rewrap` like `encrypt` and `decrypt`

Request and response match existing batch decrypt and encrypt one respectivly

PS: Manual implementation is ugly because `VaultTransitTemplace.vaultOperations` and `toEncryptionResults` are private

PR #819, opened 2023-09-29
Add transit rewrap batch method

nbaars:
Closes: gh-687


mp911de:
Thank you for your contribution. That's merged and polished now. I provided also a reactive variant of `rewrap`.

Issue #477, opened 2019-09-27
Improvement: Enhance PKI API - reading and writing certificates


Vity01:
It looks like the implementation support is far behind current Hashicorp Vault functionality. 
Eg. see [https://www.vaultproject.io/api/secret/pki/index.html#read-ca-certificate](https://www.vaultproject.io/api/secret/pki/index.html#read-ca-certificate)
Also the current functionality of generateCertificates is not feature complete - 
see [https://www.vaultproject.io/api/secret/pki/index.html#generate-certificate](https://www.vaultproject.io/api/secret/pki/index.html#generate-certificate)
the server response contains property "ca_chain" - at this time it's not possible to read it. 

Sadly there is no way how to bypass missing implementation. Eg.:
```
final VaultResponseSupport<byte[]> read = vaultOperations.read("/pki/ca/pem", byte[].class);
```
will throw 
```
.RestClientException: Could not extract response: no suitable HttpMessageConverter found for response type [org.springframework.vault.client.VaultResponses$1@5c6fae3c] and content type [application/pkix-cert]
```
and there is no easy way to add custom implementation of HttpMessageConverter  into the rest client, since the only 3 converters are hardcoded (why not use the app context default ones?) - see `org.springframework.vault.client.VaultClients#createRestTemplate()`

mp911de:
We haven't updated the PKI functionality for quite a while. Feel free to submit a pull request.

Vity01:
Do you see any workaround how to read `application/pkix-cert` response from the Vault? 
I don't see any. It's impossible to reuse or to touch the RestTemplate instances inside of `VaultTemplate` in the clean way. I guess the resttemplates/webclients clients should be converted into beans to be able to adapt them. Static generations inside of `VaultClients` don't help...
Also it's practically impossible to create own instance of VaultTemplate without massive code copying and hacking/coding. :-(

mp911de:
`VaultTemplate` exposes callback methods `doWithVault` and `doWithSession(…)` (invocation without/with a session token) to expose a `RestTemplate` associated with the Vault client.

Have you tried the following:
 
```java
byte[] bytes = vaultOperations.doWithSession(restOperations -> restOperations
		.getForObject("/pki/ca/pem", byte[].class));
```

Reading a `VaultResponseSupport<byte[]>` isn't possible as `VaultResponse` expects the standard-response format which is JSON carrying `auth`, `data` and other fields.

[...]

PR #820, opened 2023-09-29
Update PKI Vault operations

nbaars:
Add some extra fields through the request. The `userIds` one depends on a later Vault [version](https://github.com/hashicorp/vault/releases/tag/v1.14.2), those are behind `RequiresVaultVersion`. 

Related to: gh-477

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
@@ -113,9 +113,8 @@ public interface VaultPkiOperations {
 	 * Retrieves the specified issuer's certificate. Includes the full ca_chain of the
 	 * issuer.
 	 * @param issuer reference to an existing issuer, either by Vault-generated
-	 * identifier, or the name assigned to an issuer. If empty or {@literal null}, the
-	 * issuer will default to the literal string default to refer to the currently
-	 * configured default issuer.
+	 * identifier, or the name assigned to an issuer. Pass the literal string 'default' to
+	 * refer to the currently configured issuer.
 	 * @return the {@link VaultIssuerCertificateRequestResponse} containing a
 	 * {@link org.springframework.vault.support.Certificate}
 	 * @see <a href=
@@ -123,7 +122,7 @@ public interface VaultPkiOperations {
 	 * /pki/issuer/:issuer_ref/json</a>
 	 *
 	 */
-	VaultIssuerCertificateRequestResponse getIssuerCertificate(@Nullable String issuer) throws VaultException;
+	VaultIssuerCertificateRequestResponse getIssuerCertificate(String issuer) throws VaultException;
 
 	/**
 	 * Retrieves the specified issuer's certificate. Includes the full ca_chain of the
@@ -134,6 +133,6 @@ public interface VaultPkiOperations {
 	 * "https://www.vaultproject.io/api/secret/pki/#read-issuer-certificate">GET
 	 * /pki/issuer/:issuer_ref/{der, pem}</a>
 	 */
-	InputStream getIssuerCertificate(@Nullable String issuer, Encoding encoding) throws VaultException;
+	InputStream getIssuerCertificate(String issuer, Encoding encoding) throws VaultException;
 
 }
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
@@ -22,6 +22,7 @@ import org.springframework.vault.VaultException;
 import org.springframework.vault.support.CertificateBundle;
 import org.springframework.vault.support.VaultCertificateRequest;
 import org.springframework.vault.support.VaultCertificateResponse;
+import org.springframework.vault.support.VaultIssuerCertificateRequestResponse;
 import org.springframework.vault.support.VaultSignCertificateRequestResponse;
 
 /**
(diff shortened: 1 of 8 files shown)


spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
```diff
@@ -108,4 +109,31 @@ enum Encoding {
...
+	 * issuer will default to the literal string default to refer to the currently
+	 * configured default issuer.
+	 * @return the {@link VaultIssuerCertificateRequestResponse} containing a
+	 * {@link org.springframework.vault.support.Certificate}
+	 * @see <a href=
+	 * "https://www.vaultproject.io/api/secret/pki/#read-issuer-certificate">GET *
+	 * /pki/issuer/:issuer_ref/json</a>
+	 *
+	 */
+	VaultIssuerCertificateRequestResponse getIssuerCertificate(@Nullable String issuer) throws VaultException;
```
mp911de, line 126: I'm not sure that this argument being `@Nullable` is a good design. I understand the background, however, the defaulting to the default issuer could be handled directly by the caller, effectively us reducing the complexity.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
```diff
@@ -108,4 +109,31 @@ enum Encoding {
...
+	/**
+	 * Retrieves the specified issuer's certificate. Includes the full ca_chain of the
+	 * issuer.
+	 * @return {@link java.io.InputStream} containing the encoded certificate or
+	 * {@literal null}
+	 * @see <a href=
+	 * "https://www.vaultproject.io/api/secret/pki/#read-issuer-certificate">GET
+	 * /pki/issuer/:issuer_ref/{der, pem}</a>
+	 */
+	InputStream getIssuerCertificate(@Nullable String issuer, Encoding encoding) throws VaultException;
```
mp911de, line 137: We should not allow `null` values for the `issuer`. See the comment above for more details.

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
```diff
@@ -63,13 +70,14 @@ public class Certificate {
 	 * @param issuingCaCertificate must not be empty or {@literal null}.
 	 * @return the {@link Certificate}
 	 */
-	public static Certificate of(String serialNumber, String certificate, String issuingCaCertificate) {
+	public static Certificate of(String serialNumber, String certificate, String issuingCaCertificate,
```
mp911de, line 73: Instead of updating existing factory methods, let's introduce new ones, please. Otherwise, we introduce a breaking change. We generally look for a smooth upgrade experience across minor versions.

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
```diff
@@ -161,4 +187,29 @@ static List<X509Certificate> getCertificates(String certificates) throws Certifi
 		return result;
 	}
 
+	/**
+	 * Retrieve the issuing CA certificates as list of {@link X509Certificate}.
+	 * @return the issuing CA {@link X509Certificate}.
+	 * @since 2.3.3
+	 */
+	public List<X509Certificate> getX509IssuerCertificates() {
```
mp911de, line 195: Note that this change is source-compatible, yet binary breaking and it will require users to recompile their code.

spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
```diff
@@ -87,18 +84,19 @@ public class CertificateBundle extends Certificate {
 	 * @param certificate must not be empty or {@literal null}.
 	 * @param issuingCaCertificate must not be empty or {@literal null}.
 	 * @param privateKey must not be empty or {@literal null}.
+	 * @param revocationTime the revocation time.
 	 * @return the {@link CertificateBundle} instead.
 	 */
 	public static CertificateBundle of(String serialNumber, String certificate, String issuingCaCertificate,
-			String privateKey) {
+			String privateKey, long revocationTime) {
```
mp911de, line 91: We should introduce a new static factory method instead of updating the existing one.

mp911de (changes requested):
I reviewed your changes and left a few comments.

spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
nbaars: I was a bit in doubt and I agree better to move this to the caller. Adjusted the code accordingly

spring-vault-core/src/main/java/org/springframework/vault/core/VaultPkiOperations.java
nbaars: Updated the code

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
nbaars: @mp911de I thought this was allowed as as there is a matching signature and return type declared in a superclass now, but not sure. Should we introduce a different method?

spring-vault-core/src/main/java/org/springframework/vault/support/CertificateBundle.java
nbaars: Updated the code

spring-vault-core/src/main/java/org/springframework/vault/support/Certificate.java
mp911de: This was more of a fyi. For minor versions, that's okay.

mp911de:
Thanks a lot. I'll take this PR from here and include it for the 3.1 GA release.

PR #841, opened 2023-12-11
fix-comparison

refark:
Hello,

I'm using Spring Cloud AWS in prod to retrieve ROTATE credentials for connecting to the OpenSearch service, but it's not working as expected. I can see, that Spring Cloud loads new credentials every hour, but the environment variables don't seem to be updating. To achieve the desired behavior, I had to make a small, specific change. For me it seems like a small bag fix.

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/core/env/LeaseAwareVaultPropertySource.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/core/env/LeaseAwareVaultPropertySource.java
@@ -15,10 +15,7 @@
  */
 package org.springframework.vault.core.env;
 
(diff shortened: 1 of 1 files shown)


mp911de:
I wonder why the secret instance is no longer the same as the requested one. Did you find out by chance why the secret instance has changed?

refark:
No, I didn't. I have a guess. AWS credentials are requested twice at the service start. At some point, one of the credentials becomes revoked, and the app continues to work with one secret instance (based on my understanding of the logs). It's worth mentioning that we have a custom properties source locator to copy some properties from the postgres table into the environment. Maybe this is the cause of the second credentials request.

Issue #842, opened 2023-12-13
EnvironmentVaultConfiguration does not check for spring.cloud prefix by default. 


alvonellos:
I was using this in my project, and following this guide:[Getting Started]( https://spring.io/guides/gs/vault-config/) , I specified my configs as follows: 

```
spring.cloud.vault.uri=${URI}
spring.cloud.vault.token=${TOKEN:}
spring.cloud.vault.scheme=https
spring.cloud.vault.kv.enabled=true
spring.cloud.vault.kv.backend=kv
spring.cloud.vault.kv.version=v1
spring.cloud.vault.kv.profile-separator='/'
spring.cloud.vault.kv.default-context=application
spring.cloud.vault.kv.application-name=${HOSTNAME}
spring.config.import=vault://
```

And noticed that there was an issue with my configs because vaultEndpoint() inside: org/springframework/vault/config/EnvironmentVaultConfiguration.java, does not check for location spring.cloud.vault.uri, as well as vault.uri. I could set a prefix using: [Prefix ](https://www.baeldung.com/spring-boot-env-variable-prefixes), but I think it should check this by default. 

I'll submit a pull request making the change, if that's ok?

alvonellos:
Inside 	

```
@Override
	public VaultEndpoint vaultEndpoint() {

		String uri = getProperty("vault.uri");
		if (uri != null) {
			return VaultEndpoint.from(URI.create(uri));
		}

		throw new IllegalStateException("Vault URI (vault.uri) is null");
	}
```

getProperty does not check for common locations, such as spring.cloud, which is used commonly in instructions.

alvonellos:
I made the code changes, and then tested manually and then made sure that all the unit tests pass.

mp911de:
This is by design as Spring Vault is not Spring Cloud Vault and `EnvironmentVaultConfiguration` is part of Spring Vault and not the Boot-based variant.

PR #843, opened 2023-12-14
3.1.x environment vault configuration does not check for spring.cloud prefix by default 842

alvonellos:
This closes https://github.com/spring-projects/spring-vault/issues/842 and allows to search the properties for both the root path "", as well as ones prefixed by "spring.cloud." automatically.


mp911de:
As per #842, we do not want to introduce cloud prefixes.

PR #866, opened 2024-05-15
Apply read timeout to http-components request factory. Fixes #861

### LINKED ISSUE #861: Read-Timeout not applied with Apache Http Components and no-SSL
When using Apache Http Components with an http (non-SSL) Vault URI, the read-timeout option is not applied to the `HttpClientBuilder`.  This results in the Apache-default of 3-minutes always being used, regardless of the settings provided via Spring Boot.

I've created a trivial example project to demonstrate the issue here: https://github.com/p120ph37/spring-vault-test

I believe the fix would be to add `.setResponseTimeout(Timeout.ofMilliseconds(options.getReadTimeout().toMillis()))` here:
https://github.com/spring-projects/spring-vault/blob/542786151495f1bde4d51434c0067705e58f62f1/spring-vault-core/src/main/java/org/springframework/vault/client/ClientHttpRequestFactoryFactory.java#L329-L333

I have prototyped this fix in the `FixedClientFactoryWrapper` in my test suite -- test that fix by uncommenting this line: https://github.com/p120ph37/spring-vault-test/blob/0ffcea4bc2d7401721f29b419c9596c91827023b/src/test/java/spring/vault/test/TimeoutTests.java#L43

Entea:
Fixes read timeout issue - spring webflux application could hang indefinitely on read if vault is not responding :)

### DIFF
--- a/spring-vault-core/src/main/java/org/springframework/vault/client/ClientHttpRequestFactoryFactory.java
+++ b/spring-vault-core/src/main/java/org/springframework/vault/client/ClientHttpRequestFactoryFactory.java
@@ -328,6 +328,7 @@ public class ClientHttpRequestFactoryFactory {
 
 			RequestConfig requestConfig = RequestConfig.custom()
 				.setConnectTimeout(Timeout.ofMilliseconds(options.getConnectionTimeout().toMillis()))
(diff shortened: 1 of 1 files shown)


mp911de:

PR #890, opened 2024-12-20
Adding leading slash to prevent URI is not absolute errors

moin-oss:
The lack of a leading slash here results in a java.lang.IllegalArgumentException: URI is not absolute error when using AwsIamAuthentication createTokenUsingAwsIam method


moin-oss:
As it turns out, this is only an issue when using only the base Spring RestTemaplate, and can be resolved by using Vault client's RestTemplateBuilder.  Closing this as not needed.

PR #906, opened 2025-02-26
Correct error message when failing to unwrap AppRole

mamachanko:
Signed-off-by: Max Brauer <mbrauer@vmware.com>


mp911de:
Thank you for your contribution. That's merged and backported now.

PR #958, opened 2025-11-24
VaultPropertySource additonal property transformers

LeMakhno:


spring-vault-core/src/main/java/org/springframework/vault/annotation/VaultPropertySource.java
```diff
@@ -113,6 +114,12 @@
 	 */
 	Renewal renewal() default Renewal.OFF;
 
+    /**
+     * Specify additional property transformer classes
+     * {@link org.springframework.vault.core.util.PropertyTransformer}.
+     */
+    Class<? extends PropertyTransformer>[] propertyTransformers() default {};
```
mp911de: Introducing another flavor of property transformation creates ambiguity and takes away the simplicity of use. We have property-prefixing in place and we don't want to give up on simplicity.

The huge benefit of the current design approach is that all property transformations are visible from just looking at the annotation. Any other transformers require navigation and indirection. 

Instead, I suggest exploring a design towards annotation-based property mapping. Admittedly, `@VaultProperties(mapping = {@PropertyMapping(from="password", to="spring.datasource.password")}` reads rather clunky, but that is still a direction worth exploring. Maybe you want to explore textblocks or other syntactical variants so that we come to a design that is simple and allows for addressing your use-case.

LeMakhno: @mp911de i liked the approach with `@PropertyMapping` as annotation parameter, also i came up with idea of specitying an array of strings, where each 2 items is pair of from-to property names. But i like the variant with `@PropertyMapping` more

`
@VaultProperties(mapping = {
  "url", "spring.datasource.url",
  "user", "spring.datasource.username",
  "pwd", "spring.datasource.password"
})
`

mp911de:
Good catch about `addBootstrapRegistryInitializer` vs. `addBootstrapper`, we need to fix that in our documentation. Right now, the customizer is being called when importing Vault through the config data API by setting e.g. `spring.config.import=vault:` as it serves primarily as configuration hook for mounting secret backends available on the class path (e.g. databases, key/value). These can be disabled via `SecretBackendConfigurer.registerDefaultKeyValueSecretBackends(false)` for selective imports.

